PR #42, opened 2023-03-16
fix: turn off hotcorners and set 4 desktops

castrojo:


etc/dconf/db/local.d/01-ublue
```diff
@@ -19,19 +20,19 @@ document-font-name="Ubuntu 12"
...
 
 [org/gnome/desktop/sound]
-allow-volume-above-100-percent="true"
+allow-volume-above-100-percent=true
 theme-name="Yaru"
 
 [org/gnome/desktop/wm/preferences]
 button-layout=":minimize,maximize,close"
-num-workspaces="4"
+num-workspaces='4'
```
bsherman, line 28: no quotes on num-workspaces, integer value

castrojo:
```suggestion
num-workspaces='4'
```

PR #60, opened 2023-03-18
fix: Update Super-E shortcut isn't working

ironman820:
This pull request is in reference to #4.
I have tested and updated the shortcut defaults to make Super + E open the home folder after a clean rebase.


castrojo (approved):
Thanks for doing the legwork on this, it was bugging me and I haven't had time to investigate. Thank You!

PR #65, opened 2023-03-19
feat: add 1password

castrojo:


p5 (commented):
One thing this is missing that you may want to add is a desktop entry.  
Current configuration requires you to run from the CLI

p5:
I think something similar to this will work:

```
[Desktop Entry]
Name=1Password
Exec=/usr/1Password/1password %U
Terminal=false
Type=Application
Icon=1password
StartupWMClass=1Password
Comment=Password manager and secure wallet
MimeType=x-scheme-handler/onepassword;
Categories=Office;
```

I've not had chance to test this yet

usr/share/applications/1password.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=1Password
+Exec=/opt/1Password/1password %U
```
p5, line 3:
```suggestion
Exec=/usr/1Password/1password %U
```

Containerfile
```diff
@@ -51,3 +51,15 @@ COPY --from=cgr.dev/chainguard/cosign:latest /usr/bin/cosign /usr/bin/cosign
...
 RUN chmod +x ./kind
 RUN mv ./kind /usr/bin/kind
+
+# Install 1Password via Tarball
+RUN curl -sSO https://downloads.1password.com/linux/tar/stable/x86_64/1password-latest.tar.gz && \
+    tar -xf 1password-latest.tar.gz && \
+    rm 1password-latest.tar.gz && \
+    mkdir -p /usr/1Password && \
+    mv 1password-*/* /usr/1Password && \
+    sh /usr/libexec/1password-after-install.sh \
```
p5, line 61:
```suggestion
    mv 1password-*/* /usr/1Password && \
    cp -r /usr/1Password/resources/icons/* /usr/share/icons && \
    sh /usr/libexec/1password-after-install.sh \
```

The icons are not currently applying to the desktop entry since the system hasn't got them in the correct location.  This worked for me.
I'm sure you could probably symlink them (or `mv`) to save some space, but I'm just copying them to reduce complexity.

marcoceppi (approved):

Arcitec:
@p5 

> Edit: The icons are not working, but testing a fix

Did you figure it out? Because this looks like a clear case of moving icons into an icon folder without updating the icon file cache afterwards.

Every icon folder contains an index file which lists all icons and their internal paths for easy lookups.

The correct way to trigger a refresh is to touch the folder timestamp and then tell it to make a new cache. Here is an example for Fedora Workstation.

```sh
sudo touch /usr/share/icons/hicolor ~/.local/share/icons/hicolor
sudo gtk-update-icon-cache
```

PR #67, opened 2023-03-20
chore(ci): Turn on merge queue

castrojo:


.github/workflows/build.yml
```diff
@@ -14,6 +14,7 @@ on:
     paths-ignore:
       - '**.md'
       - '**.txt'
+  merge_group:
```
marcoceppi, line 17:
```suggestion
  merge_group:
    types: [checks_requested]
```

marcoceppi (changes requested):

.github/workflows/build.yml
marcoceppi: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge-group

marcoceppi (approved):

PR #94, opened 2023-03-26
feat: fleek autogenerated man pages

bketelsen:
copy fleek's autogenerated man pages into the right places in the container.


marcoceppi (approved):

bketelsen:
Thanks @marcoceppi

PR #105, opened 2023-03-31
chore(ci): Move ISO generation to release please

castrojo:
Yafti solved this already so using Marco's method.


bigpod98 (approved):
Looks good

PR #110, opened 2023-03-31
revert: "chore(ci): split upload action into it's own so it can find docker"

castrojo:
Reverts ublue-os/bluefin#109


castrojo:
The missing error is from a missing file, not from docker being missing.

bigpod98 (approved):

PR #147, opened 2023-04-02
chore: remove changelog.md

castrojo:
All the practice merges ended up this file, let's blow it away to unstick release-please


bigpod98:
potentially unstick

PR #158, opened 2023-04-09
feat: enable "tap-to-click" for touchpad

xalt7x:
For some reason Fedora/GNOME doesn't enable "tap-to-click" for touchpad.    
Ubuntu has it enabled by default.    
Check it with a command:   
`gsettings get org.gnome.desktop.peripherals.touchpad tap-to-click`


bigpod98 (approved):
I think thats a sane default to add

PR #163, opened 2023-04-10
fix: Apply custom background to dark mode

p5:
When you switch your global theme to dark mode, the desktop background is reset to the default GNOME background.

Since we already have this nice, dark background already used for light mode, we might as well make it the default for dark mode too.


castrojo (approved):
This looks great, and the kind of polish we need, thanks!

PR #165, opened 2023-04-10
fix: Create gnome settings background entry

p5:
If a user changes their background off of the default Lagoon, there's no way to change it back to lagoon without one of these entries.

I've tested this on my personal image (2nd screenshot), so porting the changes over to this repo.

Before:
image

After (hopefully):
image


p5:
After the latest build, it works as expected
image

bketelsen:
was just looking for this today!

PR #152, opened 2023-04-04
feat: added lines of code that makes a backup of nix before it is used so user can reset nix easily

dnkmmr69420:
title says everything

PR #167, opened 2023-04-11
fix: inconsistent dir naming in PR #152

hirnidrin:
You discussed to change the initially proposed dir name from /opt/nb to /opt/nixbackup ... code got committed using both, in an inconsistent manner.


castrojo (approved):

castrojo:
Glad you're around to catch it, thanks for the save!

PR #181, opened 2023-04-19
fix: remove podman-docker from F38

castrojo:
So the build passes


bigpod98 (commented):
How come it diesnt finish with this package imcluded?

bigpod98 (approved):

PR #197, opened 2023-04-29
feat: add Containerfile.toolbox

castrojo:
Adding a more polished default terminal

### DIFF
--- a/Containerfile.toolbox
+++ b/Containerfile.toolbox
@@ -1,4 +1,5 @@
 FROM quay.io/toolbx-images/ubuntu-toolbox:22.04
+# From https://github.com/toolbx-images/images/tree/main/ubuntu/22.04
 
 LABEL com.github.containers.toolbox="true" \
       usage="This image is meant to be used with the toolbox or distrobox command" \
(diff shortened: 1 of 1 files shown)


Containerfile.toolbox
```diff
@@ -0,0 +1,23 @@
...
+RUN sed -Ei '/apt-get (update|upgrade)/s/^/#/' /usr/local/sbin/unminimize && \
+    apt-get update && \
+    yes | /usr/local/sbin/unminimize && \
+    DEBIAN_FRONTEND=noninteractive apt-get -y install \
+        $(cat extra-packages | xargs) && \
+    rm -rd /var/lib/apt/lists/*
+
+RUN rm /toolbox-packages
+
+RUN   ln -fs /bin/sh /usr/bin/sh && \
```
p5, line 18: Will adding `ln -fs /bin/sh /usr/bin/sh` mean that any shell scripts containing `#!/usr/bin/env sh` will be ran on the host instead of the container?

p5 (approved):
The build will fail in it's current state as the `toolbox-packages` file is not present, but after we merge the next few PRs, this will start building successfully

PR #199, opened 2023-04-29
chore(ci): build toolbox

castrojo:
Build a custom ubuntu toolbox every day


.github/workflows/build-toolbox.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Lowercase Registry
+        id: registry_case
+        uses: ASzc/change-string-case-action@v5
+        with:
+          string: ${{ env.IMAGE_REGISTRY }}
+
+      # Push the image to GHCR (Image Registry)
+      - name: Push To GHCR
+        uses: redhat-actions/push-to-registry@v2
+        id: push
```
p5, line 64: If we're running on `pull_request`, we should add a condition here to skip this build and the later sign step.  Otherwise builds will fail for non-contributors.

p5 (changes requested):

p5 (approved):

p5:
Oh, I didn't realise approving would merge...  I thought the failing build would block it

PR #221, opened 2023-05-09
Switch windows (not applications) with Alt+Tab

xalt7x:
GNOME has separate shortcuts for switching windows and applications.
Application switcher by default shows grouped windows from all workspaces.
Windows switcher doesn't group windows and by default shows only windows with small previews only from the current workspace.
Originally GNOME doesn't bind shortcuts to "switch-windows" so "switch-applications" are used for both Super+Tab and Alt+Tab
When user binds Alt+Tab to "switch-windows" , "switch-applications" gets unbind for some reason.
So Ubuntu settings bind both.

### DIFF
--- a/etc/dconf/db/local.d/01-ublue
+++ b/etc/dconf/db/local.d/01-ublue
@@ -29,6 +29,12 @@ button-layout=":minimize,maximize,close"
 num-workspaces='4'
 title-bar-font="Ubuntu Bold 12"
 
+[org/gnome/desktop/wm/keybindings]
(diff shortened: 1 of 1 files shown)


xalt7x:
Also it brings consistency with current version of [https://github.com/harshadgavali/gnome-gesture-improvements](Touchpad Gesture Improvements extension) which I highly recommend for laptop users!

castrojo:
Thanks for this, I had forgotten about this setup but looking forward to having it back! Feel free to keep sending in improvements!

PR #248, opened 2023-05-29
feat: add `cargo`

castrojo:
Need this for `topgrade-rs` so we can improve the `just upgrade` command.


castrojo:
```
  = note: collect2: fatal error: cannot find 'ld'
          compilation terminated.
```

Cargo fails with this, this is the usual /etc/alternatives thing, need to find a substitute.

castrojo:
Actually let me make this dx only.

PR #249, opened 2023-05-30
feat(yafti): add syncthing flatpak

castrojo:
This is a nice self contained install of syncthing which is nice and easy.


bketelsen (approved):
delicious

Issue #257, opened 2023-06-02
Move language package managers off the image


castrojo:
We've added some language things on the base image like npm, cargo, etc. These are to get people working but we need to do this the right way before F39. 

Ideally we want to present the upstream method for installing these tools. So likely `just` shortcuts to get us started, and then maybe a yafti integration point where there's a developer section for each language that you want to pull in. 

That will leave dx to be more of the heavy tools that people might want without bundling the universe. So far I have:

- [x] Python - we need pip for other stuff so that comes with it, no reason to change anything.
- [ ] golang
- [ ] cargo/rust
- [ ] homebrew
- [ ] devbox/fleek for nix
- [ ] whatever other language stacks people want

The nice thing is once we have these anyone can put them in yafti/just and it'd spread the pattern to all the derived images, which is ultimately what we want.

wwitzel3:
My Go and Rust installations are both done with homebrew. So +1 to homebrew.

castrojo:
Alright, marco figured out homebrew and will PR it, but we'll need to add these packages: 

```
rpm-ostree install diffstat doxygen patch patchutils subversion systemtap clang15-libs clang15-resource-filesystem dyninst ed elfutils-devel libserf llvm15-libs perl-File-Find perl-Git perl-TermReadKey perl-lib subversion-libs systemtap-client systemtap-devel systemtap-runtime tbb utf8proc xapian-core-libs xz-devel elfutils-debuginfod-client-devel
```

And then remove the language managers from the image and we'll have homebrew!

castrojo:
Ok so I've PRed removing them from the image and have replaced them with the deps for homebrew, this gives us three developer stories: 

- OSX devs with homebrew
- devbox/fleek for whatever people call these things now. 😄 
- and vscode+containers for linux nerds

Marco will have a follow on PR for homebrew as we need to make adjustments to the install due to `/home/jorge` vs. `/var/home/jorge` but he has a path forward on that. 

Almost done!

PR #261, opened 2023-06-07
feat(dx): move languages off the image

castrojo:
And then replace them with the minimum bare set of deps for enabling homebrew, part of #257


marcoceppi (approved):

castrojo:
I don't think it's worth trying to figure out how to bring clan back to F37, so I think we should just drop dx 37

castrojo:
Going to hold on this until homebrew lands.

castrojo:
Ok between all the other stuff we have we can go ahead and remove these.

PR #275, opened 2023-06-18
feat: add a framework specific image

castrojo:
Making an image that implements this: https://knowledgebase.frame.work/en_us/optimizing-fedora-battery-life-r1baXZh


marcoceppi (changes requested):
You'll need to add this to the build matrix for framework like dx

marcoceppi (approved):

PR #281, opened 2023-06-19
fix(framework): append kernel arguments

castrojo:


marcoceppi (changes requested):
Those quotes are not real good quotes

marcoceppi (approved):

PR #291, opened 2023-06-22
feat(dx): install devpod and devpod cli

### LINKED ISSUE #237: dx image needs devpod
devpod also looks cool, let's add it: https://github.com/loft-sh/devpod

Between this and devbox this should handle the "portable devenv" use cases.

p5:
Closes #237 

This PR installs the latest release of DevPod and DevPod CLI.
There's a known issue with DevPod where there's some permission issues, but hopefully they will be resolved soon and DevPod will work as expected.  DevPod still works using one of the other providers though - it's just the Docker provider that's broken.

On my personal image, I've installed Docker in the meantime, which works well with DevPod.  I get that we don't want Docker on these images though.

Upstream issue: https://github.com/loft-sh/devpod/issues/413

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -77,6 +77,12 @@ COPY --from=cgr.dev/chainguard/helm:latest /usr/bin/helm /usr/bin/helm
 COPY --from=cgr.dev/chainguard/ko:latest /usr/bin/ko /usr/bin/ko
 COPY --from=cgr.dev/chainguard/minio-client:latest /usr/bin/mc /usr/bin/mc
 
+# Install DevPod
+RUN wget https://github.com/loft-sh/devpod/releases/latest/download/DevPod_linux_x86_64.rpm -O /tmp/devpod.rpm && \
+  rpm-ostree install /tmp/devpod.rpm && \
+  wget https://github.com/loft-sh/devpod/releases/latest/download/devpod-linux-amd64 -O /tmp/devpod && \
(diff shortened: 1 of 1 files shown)


Containerfile
```diff
@@ -77,6 +77,12 @@ COPY --from=cgr.dev/chainguard/helm:latest /usr/bin/helm /usr/bin/helm
 COPY --from=cgr.dev/chainguard/ko:latest /usr/bin/ko /usr/bin/ko
 COPY --from=cgr.dev/chainguard/minio-client:latest /usr/bin/mc /usr/bin/mc
 
+# Install DevPod
+RUN wget https://github.com/loft-sh/devpod/releases/latest/download/DevPod_linux_x86_64.rpm -O /tmp/devpod.rpm && \
+  rpm-ostree install /tmp/devpod.rpm && \
+  wget https://github.com/loft-sh/devpod/releases/latest/download/devpod-linux-amd64 -O /tmp/devpod && \
+  install -c -m 0755 /tmp/devpod /usr/bin
```
p5, line 84: It's been a while since contributing here.  Please let me know if there's some other conventions I should be following.  It's not too clear to me when to start a new `RUN` command and when to bundle them together.

castrojo: Don't worry about this, we'll need to clean up after but let's not slow down, this is awesome, thank you!

castrojo (approved):

PR #293, opened 2023-06-25
feat(dx): open fedora userspace on shortcut

### LINKED ISSUE #265: dx should have a Fedora userspace distrobox preloaded
It would be useful to have ctrl-alt-f bring up a Fedora userspace.

p5:
Closes #265 

### NOTE: THIS IS UNTESTED!
The only thing it could break is the existing Ubuntu distrobox shortcut, but the `flatpak run` command works locally, so I don't see how it could fail.

This PR creates a new custom keybinding to open up a Fedora distrobox with BlackBox using <kbd>CTRL</kbd> + <kbd>ALT</kbd> + <kbd>F</kbd>.  
It also refactors the existing Ubuntu keybinding to call distrobox directly, rather than opening the BlackBox app and hoping the initial command remains the same.

### DIFF
--- a/dx/etc/distrobox/distrobox.ini
+++ b/dx/etc/distrobox/distrobox.ini
@@ -5,3 +5,11 @@ nvidia=false
 pull=true
 root=false
 replace=true
+
+[fedora]
+image=fedora:latest
+init=false
+nvidia=false
+pull=true
+root=false
+replace=true
--- a/dx/etc/distrobox/distrobox.ini
+++ b/dx/etc/distrobox/distrobox.ini
@@ -7,7 +7,7 @@ root=false
 replace=true
 
 [fedora]
-image=registry.fedoraproject.org/fedora-toolbox:38
+image=registry.fedoraproject.org/fedora-toolbox:latest
 init=false
 nvidia=false
 pull=true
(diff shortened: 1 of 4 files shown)


etc/distrobox/distrobox.ini
```diff
@@ -5,3 +5,11 @@ nvidia=false
...
 root=false
 replace=true
+
+[fedora]
+image=fedora:latest
+init=false
+nvidia=false
+pull=true
+root=false
+replace=true
```
p5, line 15: Not sure what to do with this.  Should I simply copy and paste this file to /etc/distrobox/distrobox.ini so it overwrites the non-dx one?  Or do we want a fedora userspace in the regular image (without the shortcut)?

etc/dconf/db/local.d/01-ublue
```diff
@@ -71,10 +66,13 @@ sort-directories-first=true
 
 [com/raggesilver/BlackBox]
 command-as-login-shell=true
-use-custom-command=true
-custom-shell-command='distrobox enter ubuntu'
```
p5, line 75: Since BlackBox will not only be used for Ubuntu, I removed this custom command.

dx/etc/distrobox/distrobox.ini
p5, line 1: Overwrites the existing distrobox.ini since we only want the fedora container in dx

castrojo (approved):

PR #304, opened 2023-06-30
feat(framework): move TLP configuration into /etc/tlp.d

190n:
This lets users adjust the configuration, since `/etc/tlp.conf` overrides anything in `/etc/tlp.d`. I also removed sections of the config that aren't changed from TLP defaults, for clarity, and added a link to the Framework support article that the settings were derived from.


castrojo (approved):
This is excellent, thanks so much! Feel free to submit what you need to get it running smoothly!

190n:
Hold on a second, I forgot I wanted to write down that these settings reduce performance.

190n:
How does that look?

PR #320, opened 2023-07-16
feat(just): add an assemble shortcut

castrojo:
This will refresh and replace all the distroboxes you've defined in /etc/distrobox/distrobox.ini


bketelsen (approved):
LGTM

PR #321, opened 2023-07-17
feat(just): add fix for fractional scaling

castrojo:


castrojo:
I'm just going to turn this on by default.

PR #341, opened 2023-07-24
fix(framework): add new boot parameter needed for 6.4.4

castrojo:


bigpod98 (approved):
Idk if this is needed just approving what jorge made :)

PR #342, opened 2023-07-24
fix(framework): pin to kernel 6.3.11-200

castrojo:


bigpod98 (approved):

bsherman (changes requested):
Suggest using 6.3.12 not 6.3.11

https://bodhi.fedoraproject.org/updates/FEDORA-2023-94ec7a8b52

bsherman:
@castrojo should rename the PR to say 6.3.12, not 6.3.11 ;-) but otherwise G2G

bsherman:
> @castrojo should rename the PR to say 6.3.12, not 6.3.11 ;-) but otherwise G2G

Oh, i can do that mysefl. :-) all good

bsherman (approved):

bsherman:
oh yeah, but it fails because of version deps :-1:

castrojo:
Yeah, a full kernel downgrade is more complex than I realized. Maybe someday it'll be this clean.

PR #346, opened 2023-07-27
feat(just): add homebrew shortcut

castrojo:


bketelsen (approved):

bketelsen (approved):

osalbahr (changes requested):
FYI, you might want to add `NONINTERACTIVE=1` just in case (as in https://github.com/homebrew/install).

PR #358, opened 2023-07-31
fix(dx): turn on gpgcheck for vscode repo

castrojo:


p5 (approved):
Builds have passed, so shouldn't break anything

PR #366, opened 2023-08-04
fix: remove gnome-software-rpm-ostree

castrojo:
We can finally remove this again, gnome-software renders the store stuff just fine now without this. Goodbye!

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -34,18 +34,9 @@
             "all": [
                 "firefox",
                 "firefox-langpacks",
+	        "gnome-software-rpm-ostree", 
                 "gnome-tour"
             ]
-        }
-    },
-    "37": {
-        "include": {
-            "all": [
-	        "podman-docker"
-	    ]
-        },
-        "exclude": {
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -34,20 +34,11 @@
...
                 "firefox-langpacks",
+	        "gnome-software-rpm-ostree", 
                 "gnome-tour"
             ]
         }
     },
-    "37": {
-        "include": {
-            "all": [
-	        "podman-docker"
```
p5, line 44: I must be missing something.  How does `podman-docker` relate to Gnome Software?

I understand why we're removing it from all images - because we're going to install the UBlue version, but what was the issue with podman-docker?

p5: Oh, F37 builds are disabled anyway.  This has no effect.

LGTM

p5 (approved):

p5:
DevPod RPM package failed to download.  Looks like a network issue so retrying

PR #387, opened 2023-08-08
feat: add kns and kctx tools to bluefin

jeefy:
Need to find a clean way to also include completion for bash/zsh but that could be a userland/`just` follow-up :)

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -88,6 +88,13 @@ RUN rpm-ostree install $(curl https://api.github.com/repos/loft-sh/devpod/releas
   wget https://github.com/loft-sh/devpod/releases/latest/download/devpod-linux-amd64 -O /tmp/devpod && \
   install -c -m 0755 /tmp/devpod /usr/bin
 
+# Install kns/kctx and add completions for Bash
+RUN wget https://raw.githubusercontent.com/ahmetb/kubectx/master/kubectx -O /usr/bin/kubectx && \
+    wget https://raw.githubusercontent.com/ahmetb/kubectx/master/kubens -O /usr/bin/kubens && \
+    chmod +x /usr/bin/kubectx /usr/bin/kubens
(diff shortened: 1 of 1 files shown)


castrojo (approved):

Containerfile
```diff
@@ -88,6 +88,11 @@ RUN rpm-ostree install $(curl https://api.github.com/repos/loft-sh/devpod/releas
   wget https://github.com/loft-sh/devpod/releases/latest/download/devpod-linux-amd64 -O /tmp/devpod && \
   install -c -m 0755 /tmp/devpod /usr/bin
 
+# Install kns/kctx and add completions for Bash
+RUN git clone https://github.com/ahmetb/kubectx /opt/kubectx && \
+    ln -s /opt/kubectx/kubectx /usr/local/bin/kctx && \
+    ln -s /opt/kubectx/kubens /usr/local/bin/kns
```
castrojo, line 94:
```suggestion
    ln -s /opt/kubectx/kubectx /usr/bin/kctx && \
    ln -s /opt/kubectx/kubens /usr/bin/kns
```

In this context that directory is for you when you're consuming on your local image, so it'll fail if we try to put stuff in /usr/local at the build level.

castrojo (changes requested):

castrojo (approved):

castrojo (approved):

castrojo:
Looks like the latest versions have been ported to golang and are available on the releases page: https://github.com/ahmetb/kubectx/releases

For that I would just copy the pattern earlier in the containerfile that we use for kind to get the binaries that way. And then just git clone the completion directory wherever in /usr/share they're supposed to go: https://github.com/ahmetb/kubectx/tree/master/completion

castrojo (approved):

PR #395, opened 2023-08-10
feat: adding framework wallpaper and directory info

storyaddict:
Resolves this desired change https://github.com/ublue-os/bluefin/tree/main/framework/usr/share/wallpapers by @castrojo 

Adds copy command for wallpaper location and renames folder `/framework/usr/share/wallpapers` TO `framework/usr/share/backgrounds` to match naming convention.


castrojo:
Close, but I think this is what we want is to just rename `framework/usr/share/wallpapers` to `framework/usr/share/backgrounds`  and then everything works out. 😄

storyaddict:
Ah, I see where my mistake was.

castrojo (approved):

PR #396, opened 2023-08-11
feat: make power button be a power button

castrojo:


bketelsen (approved):
Happy Birthday Dustin!

PR #397, opened 2023-08-11
feat(just): add JetBrains Toolbox app install

### LINKED ISSUE #393: Just shortcut for jetbrains toolbox
This would be a great convenience shortcut for this audience:

https://www.jetbrains.com/toolbox-app/

and then that app manages the installation and updating of all the jetbrains tools in the user's space. Easy win. 😄

bobslept:
Fixes #393

### DIFF
--- a/usr/share/ublue-os/just/custom.just
+++ b/usr/share/ublue-os/just/custom.just
@@ -81,6 +81,16 @@ gnome-extensions:
   gext install pano@elhan.io
   gext install weatheroclock@CleoMenezesJr.github.io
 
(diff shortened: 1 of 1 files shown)


castrojo:
Hmm, no way to snag then thing without us having to update that sha when they update?

castrojo:
Yeah if we could figure out how to ingest that then that's one less thing to remember and would hopefully future proof us?

bobslept:
What do you think of this? Now we only need to change BUILD_VERSION to match the latest.

castrojo (approved):

castrojo:
Yeah that's much easier, we can always improve later, thanks!

PR #403, opened 2023-08-12
feat(just): add devmode-on/off

### LINKED ISSUE #274: Make a Developer mode toggle
aka. turbo button. I think a gui app that just let's you toggle it would be awesome but we could start with a quick set of just aliases, something like:

- `just devmode-on`
- `just devmode-off`

Then just switch between the images. If someone wants to work on this I think it'd be great!

bobslept:
Fixes #274 

This checks the current booted image. Based on that it let you switch to a dx variant or back to normal with:
```
just devmode-on
just devmode-off
```

Tested on all the images. 
```
bluefin
bluefin-dx
bluefin-nvidia
bluefin-dx-nvidia
bluefin-framework
bluefin-dx-framework
```

### DIFF
--- a/usr/share/ublue-os/just/custom.just
+++ b/usr/share/ublue-os/just/custom.just
@@ -59,6 +59,32 @@ cockpit:
 code-profile:
   xdg-open https://vscode.dev/profile/github/c761b7738e9a7b02286d6d94cb2d1ecd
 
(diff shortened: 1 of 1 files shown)


castrojo (approved):
We've needed this for a while, thank you!

castrojo (approved):

PR #404, opened 2023-08-12
feat: add descriptions to just commands

castrojo:


usr/share/ublue-os/just/custom.just
```diff
@@ -122,35 +131,37 @@ jetbrains-toolbox:
...
 
+# Install nix and Devbox (Global Profile)
 nix-devbox-global:
   echo 'Installing devbox global profile.'
   devbox global pull https://devbox.getfleek.dev/high
   echo 'run "devbox global run install-bash-hook" to configure bash shell'
   echo 'run "devbox global run install-zsh-hook" to configure zsh shell'
   echo 'run "devbox global run" to see other available configuration commands'
 
-tea:
```
bobslept, line 139: Was this intended?

bobslept (commented):
Very nice! Left a small note on the `tea` command to be sure.

usr/share/ublue-os/just/custom.just
castrojo: Yeah I tried it for a bit but feels like we have more established things in here. We could leave it, just feels like it's not really picking up steam?

bobslept: Let's clean it up then

castrojo:
Woo! Got them in config as well, we should be nice and descriptive across the board! https://github.com/ublue-os/config/pull/94

bsherman (approved):
Looks good to me :-)

PR #422, opened 2023-08-16
feat: add mesa-libGLU

castrojo:
So that davinci resolve can work


castrojo:
Actually let's hold off, I want to investigate a containerized way to do this.

PR #423, opened 2023-08-17
fix: add back a distrobox.conf

castrojo:
This should ensure that the right image is pulled on initial usage


bobslept (approved):
Awesome, works great on my end.

PR #424, opened 2023-08-17
fix: podmansh improvements

jeefy:
- passes in passwd/group
- passes in selinux
- adds passwordless sudo to the toolbox
- adds a "stop" command when running `just podmansh` in case it was already running (should only run into this while debugging or messing around, not really common)

### DIFF
--- a/usr/share/ublue-os/just/custom.just
+++ b/usr/share/ublue-os/just/custom.just
@@ -170,6 +170,30 @@ podmansh-switch IMAGE:
   echo "Image now switched to {{IMAGE}}"
 
 # Install better touch-friendly GNOME extensions
+podmansh:
+  sudo mkdir -p /etc/containers/systemd/users/${UID}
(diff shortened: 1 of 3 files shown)


usr/share/ublue-os/just/custom.just
```diff
@@ -169,7 +170,6 @@ podmansh-switch IMAGE:
   systemctl --user start podmansh.service
   echo "Image now switched to {{IMAGE}}"
 
-# Install better touch-friendly GNOME extensions
```
bobslept, line 172: Lets leave in the just description for `just touch` command

bobslept (commented):
Cool!

usr/share/ublue-os/just/custom.just
castrojo: I'll add this back in in a new PR

castrojo (approved):

PR #426, opened 2023-08-17
fix(just): devbox global shell hooks

fossrob:
When running `just nix-devbox-global` it echoes the commands to run to configure your `bash` or `zsh` shell but the devbox scripts are misnamed: `bash/zsh` < - > `hook`

```
$ devbox global run install-bash-hook
/var/home/robin/.local/share/devbox/global/default/.devbox/gen/scripts/.cmd.sh: line 3: install-bash-hook: command not found

$ devbox global run install-zsh-hook
/var/home/robin/.local/share/devbox/global/default/.devbox/gen/scripts/.cmd.sh: line 3: install-zsh-hook: command not found

```


bobslept (approved):
Good catch!

PR #434, opened 2023-08-22
fix: move /etc/ to /usr/etc/

### LINKED ISSUE #428: Move most customizations to /usr/etc
We should be shipping things in /usr/etc so that the end user has a safe fallback when customizing their installs. 

Then we document how to do that. This is something I should have done before it got out of hand and then it got out of hand, heh.

bobslept:
Fixes #428 

This moves all the `/etc/` directories to their corresponding `/usr/etc` directory. We still need copy back a few files to make the builds work. Like systemd service files or repo files. 

I hope this is the correct approach.


castrojo:
This looks good to me but considering the change let's get a few more eyeballs on this. Thanks!

bketelsen:
Looks good to me too. @castrojo I'm good with merging

castrojo (dismissed):

bketelsen:
these build errors are from the mutter-vrr package, unrelated to this PR

Containerfile
```diff
@@ -10,8 +10,9 @@ FROM ${BASE_IMAGE}:${FEDORA_MAJOR_VERSION} AS bluefin
 ARG IMAGE_NAME="${IMAGE_NAME}"
 ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION}"
 
-COPY etc /etc
 COPY usr /usr
+COPY usr/etc/systemd /etc/systemd
+COPY usr/etc/yum.repos.d /etc/yum.repos.d
```
p5, line 15: Would it be worth keeping both the `/etc/` and `/usr/` files in the repo for easier management and understanding?

We're chucking files into `/usr/etc` in the repo but that's not where they end up in the image.  It could be confusing when debugging issues

bobslept: I think the idea of having them in `/usr/etc` is that we can reset for a fresh start. So they have to be there for that operation. When you have them in `/etc/` in the repo, changes may be that they don't end up in `/usr/etc/`. 

They are available from that location on boot, but while building the container they are not. That is why I copy them in there explicit.

I will let you all decide what is best, because I am not comfortable enough to make a good descision on this point.

p5: Jorge and others are happy with the way you've proposed, so I'm ok with it.  

I just wasn't too sure of how this would be used.

bobslept: Maybe this [ostree deployment](https://ostreedev.github.io/ostree/deployment/) is where the idea also came from. In cominbation with the config reset to get a fresh start.

> Besides the exceptions of /var and /etc then, the rest of the contents of the tree are checked out as hard links into the repository. It’s strongly recommended that operating systems ship all of their content in /usr, but this is not a hard requirement.

castrojo: Since the build is busted anyway, would it help if we document how it should be used? I'm thinking something like "we'll ship the configs in /usr/etc so you have a copy, here's how you override a few of them, like your distrobox.ini, etc?"

castrojo:
I'll work on document this, holding this until then (or someone else beats me to it).

castrojo:
https://github.com/ublue-os/website/pull/404

Containerfile
bsherman: > Maybe this [ostree deployment](https://ostreedev.github.io/ostree/deployment/) is where the idea also came from. In cominbation with the config reset to get a fresh start.

I think the more relevant part of the link is: https://ostreedev.github.io/ostree/deployment/#contents-of-a-deployment

> The deployment should not have a traditional UNIX /etc; instead, it should include /usr/etc. This is the “default configuration”. When OSTree creates a deployment, it performs a 3-way merge using the old default configuration, the active system’s /etc, and the new default configuration. In the final filesystem tree for a deployment then, /etc is a regular writable directory.


TLDR: The idea of this PR seems good, but after moving files (in repo) from `etc` to `usr/etc`, I think it's contrary to the goal to copy any files back to `/etc/*` during the build. 

Specifically I'm referring to lines like:
```
COPY usr/etc/systemd /etc/systemd
COPY dx/usr/etc/yum.repos.d /etc/yum.repos.d
COPY framework/usr/etc/systemd /etc/systemd
```

bobslept: So maybe after all the best way for now is just leave the repo structure in place and just add a `COPY /etc /usr/etc` instead like @p5 suggested? This way we don't have to think which file goes where when doing both `/etc` and `/usr/etc` at repo level. 

Or maybe even ignore this PR all together because it seems that the files we never put in `/usr/etc/` at this moment are all there to have the original version of the file. Which was our first reason for doing this if i'm correct.

castrojo:
OK what am I doing here? Do we have concensus? 😄 I've PR'ed the docs and that's merged.

bsherman:
> OK what am I doing here? Do we have concensus? 😄 I've PR'ed the docs and that's merged.

I can make some specific requests to change the PR in a way which is compatible with what has been discussed and you put on the website.

Containerfile
```diff
@@ -10,8 +10,9 @@ FROM ${BASE_IMAGE}:${FEDORA_MAJOR_VERSION} AS bluefin
 ARG IMAGE_NAME="${IMAGE_NAME}"
 ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION}"
 
-COPY etc /etc
 COPY usr /usr
+COPY usr/etc/systemd /etc/systemd
```
bsherman, line 14: Please remove these 2 lines:

```
COPY usr/etc/systemd /etc/systemd
COPY usr/etc/yum.repos.d /etc/yum.repos.d
```

Containerfile
```diff
@@ -51,8 +52,8 @@ ARG IMAGE_NAME="${IMAGE_NAME}"
 ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION}"
 
 # dx specific files come from the dx directory in this repo
-COPY dx/etc /etc
 COPY dx/usr /usr
+COPY dx/usr/etc/yum.repos.d /etc/yum.repos.d
```
bsherman, line 56: Please remove this line:
```
COPY dx/usr/etc/yum.repos.d /etc/yum.repos.d
```

Containerfile
```diff
@@ -112,8 +113,8 @@ RUN ostree container commit
 # Image for Framework laptops
 FROM bluefin AS bluefin-framework
 
-COPY framework/etc /etc
 COPY framework/usr /usr
+COPY framework/usr/etc/systemd /etc/systemd
```
bsherman, line 117: Please remove this line:
```
COPY framework/usr/etc/systemd /etc/systemd
```

bsherman (changes requested):
I have 3 comments which are requested changes which should bring things in line with what I believe was discussed and which was published to the website.

Also, this needs to be updated from main branch.

Then I think should be good to go.

bobslept:
Oh boy I think officially messed everything up now. So sorry :cry:

bsherman:
> Oh boy I think officially messed everything up now. So sorry 😢

No worries! With git and pull requests, there is rarely an unrecoverable issue.

I have a suggestion for the current situation though.  Since your PR included moving a lot of files, it's harder to merge in changes made to those files.


I suggest making a fresh new fork of `bluefin` repo. Immediately make the needed changes and create the PR, but also link to this one in the comment so all the lovely history is connected. :-)

I'll try to approve quickly assuming it's what I expect from the last set of comments here.

bobslept:
I've created a fresh one. Lets continue this PR in #441

PR #441, opened 2023-08-27
fix: move /etc/ to /usr/etc/

### LINKED ISSUE #428: Move most customizations to /usr/etc

bobslept:
Fixes #428 
Follow up of PR #434 

This moves the current `/etc/` folders to their corresponding `/usr/etc` folder.


bobslept:
All builds are failing on installing `tailscale`. The repository file is here `usr/etc/yum.repos.d/tailscale.repo`. But not is not found after moving it to the new location. 

I expect the same will happen for `code` on the `bluefin-dx` with repo file located `dx/usr/etc/yum.repos.d/vscode.repo`

bsherman:
> All builds are failing on installing `tailscale`. The repository file is here `usr/etc/yum.repos.d/tailscale.repo`. But not is not found after moving it to the new location.

Ah! I see! I'm sorry, I misunderstood the cause of this error when I saw all the out of sync changes on the other PR.

> As we don't want to copy back stuff from `/usr/etc` does this mean we need to leave an `etc` folder exist for things like repo files? And then leave a `COPY etc /etc` in the Containerfile? Or is there another option.


So, yeah, basically I completely missed the relevance of the tailscale.repo  and vscode.repo files.  I'm sorry, I think I introduced confusion here.


After rereading all the discussions, I think the simplest solution, minimal changes to accomplish the goal is as follows:

1) Update the PR to restore  `usr/etc/repos.d` to `etc/repos.d` and `dx/usr/etc/repos.d` to `dx/etc/repos.d` as those never made it into the final image anyway, given the `Containerfile` was already deleting `tailscale.repo` and `vscode.repo` .
2) Leave all other moved files in `usr/etc` as they DO end up in the image
3) This will require restoring the Containerfile lines
```
COPY etc /etc
```
and
```
COPY dx/etc /etc
```

But it's probably clearer for the intent if they are more specific like:
```
COPY etc/yum.repos.d /etc/yum.repos.d
```
and
```
COPY dx/etc/yum.repos.d /etc/yum.repos.d
```



I hope this is clearer now. And again, sorry for causing confusion.

bsherman:
Adding a summary of the discord chat @bobslept and I shared.

> bobslept: Same problem pops up with enabling a service that is not found on the new location.

I remembered we already have this problem solved for a service we install in [config](https://github.com/ublue-os/config/blob/70634bf0d9fa9d506052cc764de45d0eac59d3c0/files/usr/lib/systemd/system-preset/10-flatpak-system-update.preset) where we systemd presets.


This solution requires adding 2 new files...  probably one per service whose `.service` file is added via `usr/etc/systemd` in this repo.
And the services which would need this are `fprintd.service` (for framework) and  `dconf-update.service` (for bluefin)

bsherman (approved):
I think this should be good now!

PR #446, opened 2023-08-28
fix: enable dconf-update service via systemctl

bobslept:
Moving the service file to `usr/lib/systemd/system/dconf-update.service` we are able to enable it via `systemctl enable dconf-update.service` in the container file.  Instead of force enabling it with a symlink like we did in #442


bobslept:
The failing wolfi toolbox image is unrelated to this PR. 
In all the bluefin build logs I see the service being enabled:
` Created symlink /etc/systemd/system/multi-user.target.wants/dconf-update.service → /usr/lib/systemd/system/dconf-update.service.`

bsherman:
Yep, that does work.

I'm good with this, but if doing it this way, we should probably remove the system-preset file, too... Just gets things back to a simpler to understand default, I think.

Do you agree?

bobslept:
I am fine with that. I will remove the preset file then.

castrojo:
The wolfi check isn't required for the merge to land (only the OS images are set to required in the repo settings)

bsherman (approved):

PR #449, opened 2023-08-30
fix: remove cosign layer since main upstream provides cosign in RPM

bsherman:


castrojo:
Despite my constant complaining about RPMs I understand a good portion of our userbase likes the ability to use existing tooling. It might even be worth it just for the changelogs with the version bump being in the right list. 

You win this one. 😄

castrojo:
To add insult to injury the wolfi toolbox is broken too? 😄

castrojo (approved):

bsherman:
> Despite my constant complaining about RPMs I understand a good portion of our userbase likes the ability to use existing tooling. 

It's sigstore's fault. if they didn't ship an RPM, we'd have probably added the cosign layer to `main` upstream of `bluefin` instead of the RPM. :-)

Issue #276, opened 2023-06-18
Correctly support -dx and -framework images


castrojo:
Right now I'm inlining rpm-ostree installs into the sections for -dx and -framework. That should really be properly supported in `packages.json` so it's consistent and clean. If someone wants to take a stab at this I'd appreciate it!

PR #457, opened 2023-09-02
fix: make use of packages.json for all images

bobslept:
This uses a new introduced `ARG PACKAGE_LIST` to define which image we are building, and take the corresponding packages out of `packages.json`. Includes and excludes all work like before, just a slighty more narrowed down approach to make it work with the bluefin build proces.

Fixes: #276


bobslept:
I see warnings about: 
Warning:  one or more build args were not consumed: [IMAGE_NAME]
https://github.com/ublue-os/bluefin/actions/runs/6058404397/job/16440514675#step:8:3258

I'm not sure if that will lead to problems. Otherwise I need to add it back in. I thought it was only used in `build.sh`

castrojo (approved):
Thanks for fixing this, let's see what happens with that warning but looking like we should be fine?

bobslept:
I think we are good, did a quick skopeo compare of older images and de ones just build. All that looks good.

PR #463, opened 2023-09-03
fix(just): improve JetBrains Toolbox installation

bobslept:
Improved version of the `just jetbrains-toolbox` command. This should now always install the latest version without us having to specify it.


castrojo (approved):
Nice!!!!!!

PR #469, opened 2023-09-07
feat: Reorganizing Yafti with new submenus

Tsubajashi:
Defaults are still exactly the same, just reorganized a bit to be more readable, atleast in my opinion. Feel free to merge it or not, and if there are any questions, im always open to change things up. :)
I also added an alternative Steam Proton Updater, which is made in GTK, so it fits the desktop quite a bit better. The Tool added is here: https://flathub.org/apps/com.vysp3r.ProtonPlus

### DIFF
--- a/usr/etc/yafti.yml
+++ b/usr/etc/yafti.yml
@@ -34,8 +34,6 @@ screens:
           description: Core Applications
           default: true
           packages:
-          - Backup: org.gnome.DejaDup
-          - Blackbox Terminal: com.raggesilver.BlackBox
           - Calculator: org.gnome.Calculator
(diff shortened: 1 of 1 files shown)


castrojo (changes requested):
Thanks for this, just a few things, let's keep Blackbox in the coreutils section because we need that for the container experience. And let's just use the GTK upgrader since we default to GNOME. Everything else looks good to me, thank you!

Tsubajashi:
Does these changes look good?

bobslept (commented):
Changes look good to me

bobslept (approved):

castrojo (approved):

PR #470, opened 2023-09-07
fix(yafti): add link to announcements and discussions

bobslept:
Adds a  link to the Discussions and Announcements page on the final screen.


castrojo (approved):
Great idea!

bobslept:
Another build error from terra. That is some unstable repo lately.

PR #472, opened 2023-09-07
fix(yafti): hint on expected packages by bluefin

### LINKED ISSUE #471: How to handle disabled application in yafti, but expected by bluefin
It could happen for example that a user disables `Blackbox terminal` on first run, because why do I need another terminal? Which is totally understandable. But can lead to unexptected behavior.

The problem may occur that we inside `bluefin` point shortcuts to blackbox because we expect it to be installed, which then would not work anymore. Is this something we need to fix? And if so, what would be the best way?

bobslept:
This fixes #471 

Adds a little hint in the form of an asterix that a package is expected to be installed on a bluefin system for a complete user experience.

### DIFF
--- a/usr/etc/yafti.yml
+++ b/usr/etc/yafti.yml
@@ -31,9 +31,10 @@ screens:
       package_manager: yafti.plugin.flatpak
       groups:
         Core:
-          description: Core Applications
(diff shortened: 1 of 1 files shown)


bobslept:
I will close this for now. We go for another approach.

PR #474, opened 2023-09-07
fix(yafti): move core apps to seperate screen

### LINKED ISSUE #471: How to handle disabled application in yafti, but expected by bluefin

bobslept:
Fixes #471 

Moves the core apps to a seperate "required applications" screen like images below:
image
image

### DIFF
--- a/usr/etc/yafti.yml
+++ b/usr/etc/yafti.yml
@@ -26,7 +26,7 @@ screens:
   install-required-packages:
     source: yafti.screen.package
     values:
-      title: Installing required packages
+      title: Install required packages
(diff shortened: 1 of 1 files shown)


bobslept:
Oke let's not say `Installing required packages` but `Install required packages`

castrojo (approved):

PR #485, opened 2023-09-13
feat: install docker on bluefin-dx image

### LINKED ISSUE #483: Ship Docker
We're going to end up special casing every tool people want in order to make them work with Podman when they expect Docker. 

So let's just save ourselves a bunch of time and ship it on the -dx image. 

https://docs.docker.com/engine/install/fedora/

And then if we need to do post-config `just docker` to do the other stuff then we can do that. But ideally the least amount of friction. I think @bketelsen has some other setup things we might want to add.

p5:
Fixes #483

This PR installs Docker and related services or components.
Advice from @bketelsen was to enable IP forwarding before installing Docker Engine, so I have tried to do so here.

I am not certain about the IP forwarding stuff, so would like somebody who knows what to look out for to give it a review.


usr/share/ublue-os/just/custom.just
```diff
@@ -194,3 +194,7 @@ yafti:
 # Switch to the zsh shell
 zsh:
   sudo lchsh $USER /usr/bin/zsh 
+
+docker:
+  sudo systemctl enable --now docker
+  sudo usermod -aG docker $USER
```
p5, line 200: I use this in my image, but the contents of the PR should do the same.

```
setup-docker:
  @echo "Adding user to docker group..."
  @bash -c "sudo usermod -aG docker $(whoami)"
```

castrojo (approved):

PR #495, opened 2023-09-17
fix: add user to docker group

castrojo:
I had to log out and back in order for docker to work, I think if we add this here we should be good but I'm not sure.


p5 (approved):
This command has been hit and miss for me recently.  
Sometimes it works and others it doesn't do anything and I'm not sure why.

castrojo:
Should I perhaps leave an echo in the snippet to remind the user to log out and back in or reboot?

p5:
Let's see how well it works here.  Might just have been my installs.

PR #503, opened 2023-09-20
feat: install starship bash prompt in all images

p5:
Installs the Starship prompt by default, for all users - without needing to mess with any justfiles.  

Only possible issue is it's not easy to disable this.  
Keen to hear ideas for if/how to improve this.


p5:
This breaks global aliases.  Will investigate other ways to do this

bobslept (approved):

castrojo:
How do we turn it off? I can just add the instructions to the docs. Also thinking it might be easier long term to just have it as the default for non dx too?

bobslept:
> How do we turn it off? I can just add the instructions to the docs. Also thinking it might be easier long term to just have it as the default for non dx too?

If you comment it out / remove it from bashrc it will turn off.

p5:
> But what will happen, if it will try to load starship, after some bashrc edit from the user, which will overide the default config. You will end up with an error opening the terminal bash: starship: command not found if you switched from bluefin-dx to bluefin for example.

What I have done to solve that in another image is configure my main's `/etc/bashrc` to source all shell scripts from `/etc/bashrc.d` (similar to `profile.d`) and the eval is added to a `/etc/bashrc.d/starship.sh`.  No need for anybody to edit that one.

Thought it's too many changes to bring over here without a discussion

bobslept:
That is a neat solution, good to know! But maybe enabling it for all variants is easier in the end tho :smiley:

bobslept (approved):
Perfect solution if you ask me! :+1:

PR #522, opened 2023-09-24
feat: Allow multiple justfiles.

hirnidrin:
Proposed change:
* I can add multiple *.just files into the just/ folder, in addition to the core custom.just file.
* At build time, all *.just files are concatenated together. The new `find` command in the Containerfile makes sure that content is separated by 2 newlines, so that a possibly missing newline at the end of an individual file cannot break justfile syntax.

PS: I roll my own slightly changed image off ublue-os/bluefin. I prefer extending functionality by adding extra files, instead of mangling with existing core files. This proposed change enables that.

Usage example: https://github.com/hirnidrin/bluefin/tree/8bfb7a35cbe0503f6f8fabda581368fdd1939304/just

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -16,7 +16,7 @@ ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION}"
 ARG PACKAGE_LIST="bluefin"
 
 COPY usr /usr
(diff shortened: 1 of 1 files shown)


KyleGospo (approved):

castrojo:
Thanks!

PR #525, opened 2023-09-25
fix: tweak build.sh handling of package.json

bsherman:
The includes and excludes are handled a bit oddly in the old build.sh pattern.

This change always does the requested installs, but then always does the override remove for excludes, which enables excluding dependancies of packages which were just requested to be installed.

This should cause #519 to work as expected.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -12,24 +12,18 @@ EXCLUDED_PACKAGES=($(jq -r "[(.all.exclude | (select(.\"$PACKAGE_LIST\" != null)
                              | sort | unique[]" /tmp/packages.json))
 
 
+
+if [[ "${#INCLUDED_PACKAGES[@]}" -gt 0 ]]; then
+    rpm-ostree install \
+        ${INCLUDED_PACKAGES[@]}
+else
+    echo "No packages to install."
+fi
+
 if [[ "${#EXCLUDED_PACKAGES[@]}" -gt 0 ]]; then
     EXCLUDED_PACKAGES=($(rpm -qa --queryformat='%{NAME} ' ${EXCLUDED_PACKAGES[@]}))
 fi
-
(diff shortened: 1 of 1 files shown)


KyleGospo (approved):

bpbeatty:
I believe there are scenarios in which the old build.sh handling of packages is needed. In my fork, I install `vim-default-editor` which requires the package `nano-default-editor`to be removed beforehand. In this scenario the new build.sh fails with an error like 
```
error: Could not depsolve transaction; 1 problem detected:
 Problem: conflicting requests
  - package nano-default-editor-7.2-2.fc38.noarch from @System conflicts with system-default-editor provided by vim-default-editor-2:9.0.1429-1.fc38.noarch from fedora
  - package vim-default-editor-2:9.0.1429-1.fc38.noarch from fedora conflicts with system-default-editor provided by nano-default-editor-7.2-2.fc38.noarch from @System
  - package nano-default-editor-7.2-2.fc38.noarch from @System conflicts with system-default-editor provided by vim-default-editor-2:9.0.1927-1.fc38.noarch from updates
  - package vim-default-editor-2:9.0.1927-1.fc38.noarch from updates conflicts with system-default-editor provided by nano-default-editor-7.2-2.fc38.noarch from @System
  - package nano-default-editor-7.2-2.fc38.noarch from @System conflicts with system-default-editor provided by vim-default-editor-2:9.0.1440-1.fc38.noarch from updates-archive
```
when previously it passed with output like
```
rpm-ostree override remove nano-default-editor --install=vim-default-editor
```

bsherman:
> I believe there are scenarios in which the old build.sh handling of packages is needed. In my fork, I install `vim-default-editor` which requires the package `nano-default-editor`to be removed beforehand. In this scenario the new build.sh fails with an error like
> 
> ```
> rpm-ostree override remove nano-default-editor --install=vim-default-editor
> ```

Thanks for this comment @bpbeatty! I was thinking about this yesterday and realized we probably had a new problem, and you have described it perfectly.

I'll PR a revert, and think about how to better address my concern.

PR #528, opened 2023-09-27
fix: ensure that package override replace works properly

bsherman:
This is an addtion to PR #525 which maintains the goal of that PR, allowing for packages installed as a depdency to be later excluded, but also fixes the bug which was inadvertently created.


bsherman:
@bpbeatty this should address the bug you reported on #525

bobslept (approved):

castrojo (approved):

PR #552, opened 2023-10-01
Revert "fix: send notification on flatpak installation"

bobslept:
Reverts ublue-os/bluefin#551

Well too bad this didn't work


bobslept:
I'm not sure if we should revert. Somehow I did see the `finished flatpak installation` notification today. So it's totally functional but it seem to do something. I have no clue why I don't see the start notification, but do see the finish one.

castrojo:
Works in a VM, let's keep it!

PR #559, opened 2023-10-02
fix: disable system flathub

bobslept:
Disables the system variant of the flathub flatpak repo. This way gnome-software will always prefer user flatpaks instead of system ones.

### DIFF
--- a/usr/bin/ublue-system-flatpak-manager
+++ b/usr/bin/ublue-system-flatpak-manager
@@ -48,6 +48,7 @@ if grep -qz 'fedora' <<< $(flatpak remotes); then
   /usr/lib/fedora-third-party/fedora-third-party-opt-out
   /usr/bin/fedora-third-party disable
   flatpak remote-delete fedora --force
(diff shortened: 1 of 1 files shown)


usr/bin/ublue-system-flatpak-manager
```diff
@@ -48,6 +48,7 @@ if grep -qz 'fedora' <<< $(flatpak remotes); then
   /usr/lib/fedora-third-party/fedora-third-party-opt-out
   /usr/bin/fedora-third-party disable
   flatpak remote-delete fedora --force
+  flatpak remote-modify flathub --disable --system
```
p5, line 51: This should probably be external to this condition?
This only runs if fedora repos are present, but we want it to always run?

bobslept: It runs it once on a clean install, that's all we need right? I copy the old yafti behavior, we deleted it there instead of disabling it on the same condition.

p5 (approved):

PR #565, opened 2023-10-04
fix: Fix "Sort folders first" default in Nautilus

fiftydinar:
This additional dconf is also needed for it to work (thanks to Akzel)


fiftydinar:
Will put pull request properly with 1 commit, closing

PR #566, opened 2023-10-04
feat: Disable Gnome Software auto-updates & lock it from changing

fiftydinar:


usr/etc/dconf/db/local.d/locks/01-ublue-lock
p5, line 1: I never knew about this `/locks/` directory.  Does it work for all dconf settings?

fiftydinar: As far as I tested in my own image, everything worked, including this dconf lock.

p5: I trust you that this gnome software stuff works, I was asking in general, can anything be in this `/locks` directory where it prevents changes being made in the OS?  Or is it only select few dconf settings?

fiftydinar: Well, I'm not the expert in this regard, so I cannot clarify precisely, but I believe that anything that bluefin sets in dconf, will also work in lock too.

p5 (approved):

p5:
The failing build is unrelated to changes from this PR, so will merge.  
It's just some F39 package conflicts

ArtikusHG:
There is an issue about this in main: https://github.com/ublue-os/main/issues/194

Maybe we should consider merging this into main, and not just bluefin?

PR #580, opened 2023-10-12
feat: let's go blue

castrojo:
We might as well...


usr/etc/dconf/db/local.d/01-ublue
```diff
@@ -13,7 +13,7 @@ secondary-color='FFFFFF'
 enable-hot-corners=false
 clock-show-weekday=true
 cursor-theme="Yaru"
-icon-theme="Yaru"
+icon-theme="Yaru-blue"
 gtk-theme="Yaru"
```
bobslept, line 17:
```suggestion
gtk-theme="Yaru-blue"
```
If we go blue, lets also use the Yaru-blue gtk theme?

bobslept:
Awesome, I'm a little biased because I already use yaru-blue-(dark) myself. So I'm all in for this change.

bobslept (approved):
I'm all in.

PR #585, opened 2023-10-13
Use `usermod` instead of `lchsh` in custom.just

### LINKED ISSUE #408: `just fish` and `just zsh` failing
On a fresh rebase onto `bluefin-dx-nvidia:latest`, I can no longer `just zsh` (and also tried `just fish`)

Pretty sure it's related to https://github.com/ublue-os/bluefin/pull/402

Output: 
```
[jeefy@fedora ~]$ just zsh
sudo lchsh -s /usr/bin/zsh $USER
[sudo] password for jeefy: 
Error parsing arguments: unknown option.
Usage: lchsh [-i?] [-i|--interactive] [-?|--help] [--usage]
        [OPTION...] [user]
error: Recipe `zsh` failed on line 262 with exit code 1
```

b-:
This fixes #408 (`just fish` and `just zsh` failing). The problem (and the reason that #420 doesn't fix it) is because the `lchsh` command actually doesn't accept the new shell as a CLI argument at all!

Another way to do this (and continue using `lchsh`) would be to instead run the commands as, e.g., `echo /usr/bin/zsh | sudo lchsh $USER`, but this results in an empty (and thus slightly misleading) prompt being shown to the user:

```console
$ echo /usr/bin/zsh | sudo lchsh $USER
[sudo] password for bri: 
Changing shell for bri.
New Shell [/usr/bin/zsh]: Shell changed.
```

If we just use `usermod` it doesn't actually give any feedback, so I also threw in a little posix pipe to parse the shell back out of /etc/passwd after.

### DIFF
--- a/just/custom.just
+++ b/just/custom.just
@@ -107,7 +107,8 @@ distrobox-universal:
 
 # Switch to the fish shell
 fish:
-  sudo lchsh $USER /usr/bin/fish 
+  sudo usermod $USER --shell /usr/bin/fish 
+  printf "${USER}'s shell is now %s." "$(cat /etc/passwd | grep ":$UID:" | cut '-d:' '-f7')"
 
 # Install recommended GNOME extensions
 gnome-extensions:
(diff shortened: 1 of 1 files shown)


castrojo (approved):
This is awesome, thank you! I'm sure @bketelsen will be happy!

PR #590, opened 2023-10-13
chore: refactor chsh tasks

b-:
This adds a little shim, `/usr/bin/ujust`, to run, `just --unstable --justfile /usr/share/ublue-os/justfile`.

I think this is a much cleaner way to run our `just` tasks than putting a `~/.justfile` in the home directory, for a number of reasons. One of them is that this does not require aliasing `just=just --unstable`, and another is that a user can actually have their own `~/.justfile` with no regard for including our global one.

I didn't remove the `just` aliases from the shell dotfiles, because I'm not sure what breaking changes that might cause -- documentation, for sure, but also does yafti or anything else depend on that?

Also to be honest I'm not sure where the default just alias comes from. I didn't actually look yet :P

I did this because I realized how much duplicate code and room for errors I had in the shell change tasks I rewrote -- I think it's much better to write a task with an argument, `just chsh /usr/bin/newshell`, and then call that from other tasks that need to change the shell. However, I saw the limitations of using shell aliases to add the `--unstable` flag quite quickly -- my sub-just-invocations became longer than I'd like them to be. See https://just.systems/man/en/chapter_40.html and note how much nicer that is than `just --unstable --file /usr/share/ublue-os/justfile`...

I'd definitely like to see others' thoughts on this! Feel free to ping me on Discord @bri9.

PR is draft because I'd like others to test it just in case I messed something up here...

--bri


castrojo:
This might be more prudent to put in ublue-os/config instead of bluefin specific, nice idea!

b-:
> This might be more prudent to put in ublue-os/config instead of bluefin specific, nice idea!

Sounds like a good idea, at least for the chsh part!

b-:
This now depends on https://github.com/ublue-os/config/pull/135/checks

m2Giles:
chsh can result in unbootable systems if zsh/fish are not on a rebased image.

b-:
I wrote the chsh in ublue-os/config, and I did put an if statement in there to check if you actually passed it an executable file — otherwise it'll bail with an error. 

I definitely understand the concern — it would really suck to e.g., set your shell to `zsh` and then rebase to an image without zsh. But I'm not sure there's really such a better option.

I don't really like the idea of programmatically altering the terminal profile settings, though. That sort of thing isn't really standardized at all between any two terminals (let alone three, and you're still forgetting Konsole on Kinoite/Bazzite! What about Kitty on my own fork? Where does it end?)

More specifically, my fear is keeping track of profile schemas. If we can (as an example) set the default shell on Prompt, but then in the future Prompt changes the format/schema of the configuration without us realizing, naively editing the file anyway could break something. 

And I say that with no knowledge of how Prompt (presently) stores settings, but I'd bet anything that it's different from how Gnome Terminal and Konsole do.

m2Giles:
I think a less dangerous solution would be using a drop in file for bash shell.

With it, it would have:
`test $(command -v /SHELL/) && test -n  && exec SHELL.`

This then would basically result in a bail out if they rebase to an image without said shell. Downsides, are that it's just a file and for those that like fish/zsh they may want to remove the bash stuff.

castrojo:
Ok we've redone everything since then, thanks anyway!

PR #593, opened 2023-10-14
feat: swap out tailscale extension

### LINKED ISSUE #592: Investigate tailscale extension
https://github.com/joaophi/tailscale-gnome-qs

This one matches GNOMEs UX better, but appears to be missing the exit node functionality. We should keep an eye on it.

castrojo:
Fixes #592


bobslept (approved):

bobslept:
This seems to be only working on f39. This needs to be looked into.

PR #594, opened 2023-10-15
feat(ci): Introduce support for gts tag (tracks Fedora - 1)

EyeCantCU:
Allows users to run one release behind the latest Fedora version without any intervention every major version upgrade


castrojo (approved):
Legend, thank you!

PR #600, opened 2023-10-17
feat: add incus

### LINKED ISSUE #489: Package and ship Incus
https://stgraber.org/2023/09/13/initial-daily-packages-for-incus/

Debian can't have all the fun. 😊 We should see if someone in fedora is already working on this.

I wonder if we can just use the existing lxc lxd spec files?

castrojo:
Fixes #489

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -63,6 +63,7 @@
 				"google-droid-sans-mono-fonts",
 				"google-go-mono-fonts",
 				"ibm-plex-mono-fonts",
+				"incus",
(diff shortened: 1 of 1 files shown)


bobslept:
I would make this a draft, because there are only failed builds currently. Or did I miss something here? 
https://copr.fedorainfracloud.org/coprs/ganto/lxc4/package/incus/

bobslept (approved):

PR #605, opened 2023-10-20
feat: add gum for scripting

bketelsen:
adds `gum` from charm.sh for delightful scripting with prompts, pickers, selection boxes.


bobslept (approved):

bobslept:
Just to be sure, this is currently bluefin-dx only. Double check if that was the intention

bketelsen:
@bobslept yes, will start in dx and move up the tree after we have some experience and examples.

PR #637, opened 2023-11-03
chore: steamdeck rpm cleanup

bobslept:
I think this was not added on purpose and thus can be cleaned up. 

Or do we have the intention to ship steamdeck.rpm on our image and did I miss something?


castrojo (approved):
Vestigial limb. 😄

PR #642, opened 2023-11-06
feat(just): add fleek/homebrew shortcut

castrojo:
This will oneshot homebrew and fleek. We need to have our own profile at some point but this lets people play with it right away.

Our profile would need to be adjusted (for example git's already on the image, it doesn't need to be in here, but there's stuff on the image we might wanna move into here, like the k8s tools, etc.)


castrojo:
I get this after installation:

```
bash: zoxide: command not found
bash: direnv: command not found

```

Even though both of those are installed.

bketelsen:
> I get this after installation:
> 
> ```
> bash: zoxide: command not found
> bash: direnv: command not found
> ```
> 
> Even though both of those are installed.

Did you open a new terminal session?

castrojo:
This only happens on my laptop, my desktop seems to be working fine. Is there a way I can "reset it" to try again?

castrojo:
Ok Brian's working on this and it'll be more involved than this.

Issue #643, opened 2023-11-06
Add Development Tools to -DX


bketelsen:
### Describe the package

I'd like to request the group `Development Tools` because Linux Brew depends on them being present

### Image

All Images

PR #644, opened 2023-11-07
feat: dev tools for dx

bketelsen:
Adds the 'Development Tools' group to support Linux Brew's requirements.

Probably needs to be in Bluefin (non-dx) as well, testing in -DX first.


bketelsen:
Linux Brew requirements listed here:
https://docs.brew.sh/Homebrew-on-Linux#requirements

bketelsen:
fixes #643

p5 (approved):
As it's a requirement for Linux Homebrew, this is good to add.

bobslept (approved):

PR #664, opened 2023-11-17
chore: move to charm.sh repo

### LINKED ISSUE #661: Move to charm.sh rpm repo
https://repo.charm.sh/

This exists so we should use it instead of inlining everything. We need to move glow, gum, and add vhs back.

bobslept:
Fixes #661 

In the issue was spoken about re-addings VHS. That is not done in this PR. I would like to suggest to add VHS as part of the fleekbrew configuration.

The reasons for this is that even it is as simple as add vhs to the packages.json, it still needs a ttyd dep that we need to ship via the container. I think that would be nice to handle all via the fleekbrew.


bobslept:
I was just thinking, we want this on the base bluefin image probablly and not only in bluefin-dx because we will use them in just commands? If that is correct I need to fix this before merging.

bobslept:
I've moved them to the `bluefin` image instead of `bluefin-dx` which in my mind makes the most sense going forward.

castrojo (approved):

Issue #576, opened 2023-10-10
Recommend homebrew as the default


castrojo:
Ok so the distrobox workflow isn't working out until we get a new terminal program. I'm continually seeing people struggling with containers so I'm thinking we should just declare this as our default:

- System image: solved
- Developer workflow: devcontainers as the default (either natively in vscode or devpod), distroboxes, devbox/fleek for nix, and homebrew. (No change here) 
- CLI: Let's just use homebrew

This solves a few issues:

- It lets people install packages who don't care about containers. "I just wanted to install neotopfetch!" 
- It's self contained in `/home/linuxbrew/.linuxbrew` and can easily be blown away - doesn't interfere with the system
- It has just about everything people could want in there
  - [There are lots of linux users](https://formulae.brew.sh/analytics/os-version/90d/), not even close to the number of macos users but that's more than enough.
- It's mostly github native, there's lots of automation. [Here's dysk](https://github.com/Homebrew/homebrew-core/commit/9f836e32f53594e2d6661d300036f16425359e41) being upgraded. You can observe the packages being built in github actions and being pushed to the registry. 
- The packages are just OCI blobs stored on ghcr.io, same place the OS is, making caching locally awesome and avoiding distro mirrors entirely. [Here's the neovim](https://github.com/Homebrew/homebrew-core/pkgs/container/core%2Fneovim) package.  
- Precursory scanning of homebrew's analytics show massive usage, so we'd be using the same thing MacOS people have been using, so that's fine. 
- The downside is no ARM for linux binaries

@bketelsen's been prototyping a fleek backend so we could totally ship a default Brewfile with cool tools to supplement the starship setup, etc.

Also, getting Brew to work on silverblue was work started in this repo!

- https://github.com/ublue-os/bluefin/issues/252

But more seriously, I switched all my machines over a month ago and this is going to be what people coming from normal Linux will want: get the latest cool thing they read about on HN without caring about distro delays.

alxlg:
What does a new terminal app need for the distrobox workflow?

bigpod98:
Maybe its time i revive my old oci based package manager(never thought it could work real life but here we are)

castrojo:
> What does a new terminal app need for the distrobox workflow?

https://github.com/orgs/ublue-os/discussions/156

castrojo:
[...]

PR #669, opened 2023-11-19
Add 'brew-remove' just recipe

rohanssrao:
Runs the Homebrew uninstaller.
#576


castrojo (approved):
That was fast, thanks!

PR #673, opened 2023-11-20
feat: add just syntax checker

ArtikusHG:
Integrate our new [just syntax checker](https://github.com/ublue-os/just-action) into the build action.


bobslept (approved):

ArtikusHG:
@bobslept it seems like some builds are failing, but not because of the syntax checker, so I think it's fine to merge

bobslept:
> @bobslept it seems like some builds are failing, but not because of the syntax checker, so I think it's fine to merge

Yeah no worries. All good, thank you so much!

PR #709, opened 2023-12-04
fix(wolfi): add packages for distrobox

castrojo:
This should make distrobox launch instant


bketelsen (approved):

castrojo:
test failures are intermittent, this is fine.

PR #715, opened 2023-12-05
fix: Drop 38 support in gnome-vrr, add Xwayland patch for Nvidia 545

KyleGospo:


fiftydinar:
What about xorg-x11-proto-devel & wayland-protocols packages?

castrojo (approved):

KyleGospo:
> What about xorg-x11-proto-devel & wayland-protocols packages?

Not normally installed, if needed for other packages here they can certainly be added.

Issue #720, opened 2023-12-07
Add support for kcli


jmgilman:
### Describe the package

The [kcli](https://github.com/karmab/kcli) tool is an extremely versatile tool for working with virtual machines locally. It supports starting them from the CLI, using configuration files, layering profiles together, creating entire Kubernetes clusters, and more. It's an amazing resource for local development on VMs, and I've found the Kubernetes support is superior to things like KinD due to the ability to support much more complex setups. This seems to be in the CNCF spirit of this project :)

Luckily, `bluefin-dx` includes all of the pre-requisites already. If I understand things correctly, it should be as simple as [adding the package](https://kcli.readthedocs.io/en/latest/#package-install-method).

### Image

`bluefin-dx`

castrojo:
This seems neat, I don't have time to get to this due to travel, so whoever wants this one can snag it!

PR #724, opened 2023-12-08
feat(dx): add kcli

bobslept:
Adds the package `kcli` to the `bluefin-dx` variant.

Docs: https://kcli.readthedocs.io/en/latest/#package-install-method
Copr used: https://copr.fedorainfracloud.org/coprs/karmab/kcli/

Fixes: #720


castrojo (approved):

jmgilman:
You beat me to it. Thanks! Looks like the build failure is unrelated?

bobslept:
> You beat me to it. Thanks! Looks like the build failure is unrelated?

Oh sorry about that. Yeah the current build failures are unrelated to this PR. Some conflict in versions that are the cause of this. Just a little out of sync, one kernel 6.6.4 and some deps are still on 6.6.3. Should fix it self in the end. 

I will keep an eye on it and merge it as soon as they are green.

jmgilman:
Thanks again for pushing this through. I'm new here, so I'm not sure what the next step is. When does this show up in the latest image?

bobslept:
Should be on there on the moment it is completely merged. But it was in the progress of merging when the scheduled daily build was already started. So it missed it and produced an image without it. Let me trigger a manual rebuild.  When that is finished its on the image.  :+1:

PR #728, opened 2023-12-13
feat: Add Prompt integration

KyleGospo:
Adds Prompt Integration via nautilus-open-any-terminal, and xdg-terminal-exec.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -14,7 +14,6 @@
 				"gnome-shell-extension-gsconnect",
 				"gnome-shell-extension-tailscale-status",
 				"gnome-shell-extension-logo-menu",
-				"nautilus-open-any-terminal",
 				"hplip",
 				"ifuse",
 				"input-remapper",
@@ -108,7 +107,6 @@
 				"gnome-extensions-app",
 				"gnome-software-rpm-ostree",
 				"gnome-tour",
-				"gnome-terminal-nautilus",
 				"ublue-os-update-services"
 			],
 			"bluefin-dx": []
@@ -130,14 +128,16 @@
 	"39": {
 		"include": {
 			"bluefin": [
-				"input-leap"	
+				"input-leap",
+     "nautilus-open-any-terminal"
 			],
 			"bluefin-dx": [],
 			"bluefin-framework": []
 		},
 		"exclude": {
 			"bluefin": [
-				"podman-docker"
+				"podman-docker",
+     "gnome-terminal-nautilus"
 			],
 			"bluefin-dx": [],
 			"bluefin-framework": []
--- a/usr/etc/dconf/db/local.d/01-ublue
+++ b/usr/etc/dconf/db/local.d/01-ublue
@@ -54,7 +54,7 @@ min-alpha=0.5
 
 [org/gnome/settings-daemon/plugins/media-keys/custom-keybindings/custom1]
 binding='<Control><Alt>u'
-command='prompt -- distrobox enter ubuntu'
+command='xdg-terminal-exec "distrobox enter ubuntu"'
 name='Ubuntu Prompt'
 
 [org/gnome/settings-daemon/plugins/media-keys/custom-keybindings/custom2]
--- a/usr/etc/dconf/db/local.d/01-ublue
+++ b/usr/etc/dconf/db/local.d/01-ublue
@@ -101,6 +101,6 @@ flatpak='off'
 
 [org/gnome/shell/extensions/Logo-menu]
 menu-button-icon-image=1
-menu-button-terminal='prompt --new-window'
+menu-button-terminal='xdg-terminal-open'
 menu-button-system-monitor='flatpak run io.missioncenter.MissionCenter'
 menu-button-extensions-app='com.mattjakeman.ExtensionManager.desktop'
--- a/usr/etc/dconf/db/local.d/01-ublue
+++ b/usr/etc/dconf/db/local.d/01-ublue
@@ -52,15 +52,10 @@ customize-alphas=true
 max-alpha=0.8
 min-alpha=0.5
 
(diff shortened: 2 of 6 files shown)


packages.json
```diff
@@ -13,6 +13,8 @@
 				"gnome-shell-extension-dash-to-dock",
 				"gnome-shell-extension-gsconnect",
 				"gnome-shell-extension-tailscale-status",
+				"gnome-shell-extension-logo-menu",
```
bobslept, line 16: Is this compatible with f38 or should we maybe put this in the f39 only section? Because if we ship this from the staging copr, we probably ship a f39 (gnome 45) only version.

bobslept (commented):
Awesome, I see the fedora shortcut is still pointing to blackbox, I will push a commit for that one. 

Added a small question about compatibility of the new added extension, with the gnome 44 and 45 breaking changes we probably need a bit carefull with that.

Great work!

packages.json
KyleGospo: This is f38 compatible. this was done by building two different spec files and then being careful to make sure future builds were marked as f39 and newer only

bobslept: Perfect!

KyleGospo: Added a version gate so it'll build from the last commit that worked on GNOME 44, that also eliminates the need to be careful about build targets
https://github.com/ublue-os/Logomenu/commit/6cfa6aca55c3c6ea1f9911577123d8c9eb4ebad3

castrojo:
I noticed that prompt will not start a container on a fresh boot, so opening it leads to an error. Whose responsibility is it to ensure the container starts, the system or the terminal? 

If it's our responsibility then I think we'll need the podman autostart service and I think we can add autostart labels but we should discuss the implications of that.

KyleGospo:
@castrojo Do we want to install Prompt as a system flatpak or keep it a user flatpak like the others? I think system makes more sense given the deep integration

castrojo:
I don't have a strong opinion, whatever we think is best? Let's see what others say.

fiftydinar:
System makes more sense imo.

usr/etc/flatpak/user/install
```diff
@@ -1,6 +1,5 @@
 org.mozilla.firefox
 org.freedesktop.Platform.ffmpeg-full//22.08
-com.raggesilver.BlackBox
```
p5, line 3: :+1: Removing this line will not remove the BlackBox flatpak from existing systems, but it will mean BlackBox will not be installed by default on new installs.  
This is good behaviour.

p5 (approved):
LGTM, but would prefer waiting for someone who knows more about this flatpak than I do.

usr/etc/dconf/db/local.d/01-ublue
```diff
@@ -1,6 +1,6 @@
 [org/gnome/shell] 
-favorite-apps = ['org.mozilla.firefox.desktop', 'org.mozilla.Thunderbird.desktop', 'org.gnome.Nautilus.desktop', 'org.gnome.Rhythmbox3.desktop', 'org.libreoffice.LibreOffice.writer.desktop', 'org.gnome.Software.desktop', 'code.desktop', 'ubuntu.desktop', 'yelp.desktop']
-enabled-extensions = ['appindicatorsupport@rgcjonas.gmail.com', 'dash-to-dock@micxgx.gmail.com', 'blur-my-shell@aunetx', 'gsconnect@andyholmes.github.io']
+favorite-apps = ['org.mozilla.firefox.desktop', 'org.mozilla.Thunderbird.desktop', 'org.gnome.Nautilus.desktop', 'org.gnome.Rhythmbox3.desktop', 'org.libreoffice.LibreOffice.writer.desktop', 'org.gnome.Software.desktop', 'code.desktop', 'org.gnome.Prompt.desktop', ubuntu.desktop', 'yelp.desktop']
```
bobslept, line 2:
```suggestion
favorite-apps = ['org.mozilla.firefox.desktop', 'org.mozilla.Thunderbird.desktop', 'org.gnome.Nautilus.desktop', 'org.gnome.Rhythmbox3.desktop', 'org.libreoffice.LibreOffice.writer.desktop', 'org.gnome.Software.desktop', 'code.desktop', 'org.gnome.Prompt.desktop', 'ubuntu.desktop', 'yelp.desktop']
```
Missed a `'` here on `ubuntu.desktop`

bobslept (changes requested):
Other than a small missing character I think it looks good!

bobslept (dismissed):

KyleGospo:
If we don't want to wait on some of these improvements the dconf changes and Prompt install line can be split into their own PR, the rest of this will work today.

castrojo:
Sgtm, that way we can get one done and out of the way!

KyleGospo:
That's broken up now, see https://github.com/ublue-os/bluefin/pull/732 for the ready-to-merge parts of this.

castrojo:
I noticed in bazzite that you put it on the dock, can you put it on the dock in -dx only? we have a dir for dx dconf that should do the trick!

KyleGospo:
Yep can do, did you want layered prompt as well or did you want to wait for the flatpak?

castrojo:
Let's let it shake out for a few days in bazzite then we'll do it!

KyleGospo:
> Let's let it shake out for a few days in bazzite then we'll do it!

Re-based and added requested changes

castrojo (dismissed):
nm, let's do this.

image

KyleGospo:
@castrojo abort, need to fix one thing since this won't apply to Fedora 38. I'll re-enable shortly

KyleGospo:
@castrojo Done, fixed by only installing the Nautilus extension on 39 and pointing Logomenu to xdg-terminal-exec which is smart enough to pick the best available terminal.

castrojo (approved):

PR #738, opened 2023-12-19
feat(dx): Use DevPod from the ublue-os/staging copr

EyeCantCU:
Allows us to continue shipping an up-to-date RPM for DevPod on DX images.


p5 (approved):
Amazing!!!

PR #739, opened 2023-12-19
feat: add distrobuilder to -dx

bketelsen:
Adds distrobuilder to -dx for building custom incus and lxd images.


bketelsen:
Note: distrobuilder can't be run in a container for technical reasons, that would have been preferred.

castrojo (approved):

p5 (approved):

PR #740, opened 2023-12-20
feat(dx): turn on docker socket by default

castrojo:
And add the user to the docker group


Containerfile
```diff
@@ -135,6 +135,7 @@ RUN /tmp/build.sh && \
 
 RUN wget https://github.com/docker/compose/releases/latest/download/docker-compose-linux-x86_64 -O /tmp/docker-compose && \
     install -c -m 0755 /tmp/docker-compose /usr/bin
+    usermod -aG docker $USER
```
p5, line 138:
```suggestion
    install -c -m 0755 /tmp/docker-compose /usr/bin && \
    usermod -aG docker $USER
```

p5 (dismissed):
Not certain we can add users to groups this way, but let's give it a go

p5 (approved):

Issue #726, opened 2023-12-10
Recommend docker as the default for vscode/devcontainers


castrojo:
I switched to docker a while back so I haven't had a chance to run through a clean install lately but here's some feedback:

https://universal-blue.discourse.group/t/should-we-really-recommend-vscode-podman-as-supported/220

Maybe we should just do docker ootb? Currently we `just docker` to turn that on but maybe we should consider it having it on ootb as well, thoughts?

p5:
+1 for Docker enabled by default.  Just makes everything so much easier, and we could create one of your fancy charm just entries to switch the systemd unit to disabled

p5:
+1 for Docker enabled by default.  Just makes everything so much easier, and we could create one of your fancy charm just entries to switch the systemd unit to disabled

castrojo:
Sounds good to me, we can have it on by default and have `just docker` be like a `just devmode` toggle.

What do we do about this thing? 

https://github.com/ublue-os/bluefin/blob/main/dx/usr/etc/skel.d/.config/Code/User/settings.json

bigpod98:
Does docker not use selinux labels, i ahvent used docker on fedora cause i dont need it but based on what i know --security-opts=labels=disabled flag would still be needed besides shipping basic settings cor dev cintainers should be enough same with installing dev containers extension, been a while since i setup devcontainers since i have settings sync in vscode

bigpod98:
> Sounds good to me, we can have it on by default and have `just docker` be like a `just devmode` toggle.
> 
> What do we do about this thing? 
> 
> https://github.com/ublue-os/bluefin/blob/main/dx/usr/etc/skel.d/.config/Code/User/settings.json

Parts of this would need to be removed as it is setup for podman not docker

pid42:
> Does docker not use selinux labels, i ahvent used docker on fedora cause i dont need it but based on what i know --security-opts=labels=disabled flag would still be needed besides shipping basic settings cor dev cintainers should be enough same with installing dev containers extension, been a while since i setup devcontainers since i have settings sync in vscode
> 
> But i also use devcontainers as a way to start container and dont use more advanced features of devcontainers

It does not seem to be the case. 
I did enable docker here, removed all my 'extra' flags... Everything just works(tm). All vs code commands, generated files, simple, etc... All out of the box without "special" arguments

bigpod98:
[...]

PR #745, opened 2023-12-21
feat: add dx-specific user setup, install devcontainers/docker vscode extensions by default

jeefy:
This sets up a fresh install of Bluefin to have VSCode  set up to use a Distrobox container (via devcontainers)

There needs to be a shim in place to ensure you enter the distrobox container as the local user and not as root. 

It does require Docker to be available on the host (which is a requirement of the `devcontainers` extension

Refs:
#726 
https://universal-blue.discourse.group/t/should-we-really-recommend-vscode-podman-as-supported/220/13


bketelsen:
Is there a place to document how to specify the distrobox container in the devcontainer configs?

jeefy:
> Is there a place to document how to specify the distrobox container in the devcontainer configs?

There may be, but I don't think that's something we should set by default. Since the lifecycle of the Distrobox container is outside of VSCode, we shouldn't make assumptions on what the container is named.

p5 (commented):
This is great, although my only concern is this will try to install the VSCode extensions on all Bluefin images, whereas Bluefin DX is the only one where VSCode is installed.

Not sure what the best approach would be.  I guess conditionally running the scripts based on the contents of IMAGE_INFO JSON file

jeefy:
> I guess conditionally running the scripts based on the contents of IMAGE_INFO JSON file

Good call @p5, updated the PR with that :)

usr/bin/bluefin-dx-user-setup
castrojo, line 1: Shouldn't this be `dx/usr/bin/`?

p5: Negative.  Since this script runs on the live system, the system has no knowledge of the dx/ directory.  
The path to bash does not differ on regular and DX.

castrojo:
Any final comments on this one? I'd like to start testing this this weekend!

m2Giles:
Would this break normal dev-containers forcing the use of distrobox due to the docker path being that shim?

usr/share/ublue-os/configs/vscode/settings.json
```diff
@@ -0,0 +1,14 @@
+{
```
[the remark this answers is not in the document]
bketelsen: This should be in a separate named profile I think.
https://code.visualstudio.com/docs/editor/profiles

bketelsen (changes requested):

jeefy:
After talking with @castrojo a bit, this is the route we settled on. :) Does change the PR quite a bit but it simplifies it. 🙌

KyleGospo:
Looks good, may want to increase the version at the top of the user setup to ensure this is run for everyone

dx/usr/bin/bluefin-dx-user-setup
```diff
@@ -0,0 +1,12 @@
...
+
+# Including this check here as well as a failsafe
+if [[ -f $USER_SETUP_VER_FILE && "$USER_SETUP_VER" = "$USER_SETUP_VER_RAN" ]]; then
+  echo "User setup has already run. Exiting..."
+  exit 0
+fi
+
+# Setup VSCode
+# Pre-install preferred VSCode Extensions
+code --install-extension ms-vscode-remote.remote-containers
```
EyeCantCU, line 11: Since this would need internet access, I'd suggest creating a separate user service that runs after network-online.target otherwise the service will fail if someone boots for the first time and doesn't have an active connection

Something like this:

```
[Unit]
Description=Sets up VS Code
Wants=network-online.target
After=network-online.target bluefin-dx-user-setup.service

[Service]
Type=oneshot
ExecStart=/usr/bin/bluefin-dx-vscode
Restart=on-failure
RestartSec=30
StartLimitInterval=0

[Install]
WantedBy=default.target
```

usr/bin/ublue-user-setup
```diff
@@ -17,6 +17,13 @@ IMAGE_NAME=$(jq -r '."image-name"' < $IMAGE_INFO)
 # Enable NTP
 timedatectl set-ntp true
 
+if [[ $IMAGE_NAME = "bluefin-dx" ]]; then
```
EyeCantCU, line 20: Instead of `=`, use `=~` for comparison. Different variants will have different names. I.E. `$IMAGE_NAME` for an NVIDIA system is `bluefin-dx-nvidia`

castrojo:
Will that clobber people's existing setups? I'm fine with it being new-install only so we don't interrupt people's existing flows.

usr/bin/ublue-user-setup
```diff
@@ -17,6 +17,13 @@ IMAGE_NAME=$(jq -r '."image-name"' < $IMAGE_INFO)
...
+
+  source /usr/bin/bluefin-dx-user-setup
```
EyeCantCU, line 22: If you keep this hook, increment the script version, `USER_SETUP_VER`, so that existing users get the new stuff

EyeCantCU:
> Will that clobber people's existing setups? I'm fine with it being new-install only so we don't interrupt people's existing flows.

Unfortunately that's true. If this was its separate thing (I think it should be a separate service since it uses an internet connection to install those extensions) it wouldn't be an issue though

castrojo:
Yeah can we do that @jeefy? Sorry to make this process longer than it needs to be. (On the plus side after this we're done!)

EyeCantCU (changes requested):
Left some comments regarding converting this over to a separate service and a couple other things

jeefy:
All comments should be addressed. 

If we add additional post-install config steps, we might want to figure out a better way to track what's run. A bunch of dot files in a home feelsbadman, but that's a problem for another PR. :)

EyeCantCU:
> All comments should be addressed. 
> 
> If we add additional post-install config steps, we might want to figure out a better way to track what's run. A bunch of dot files in a home feelsbadman, but that's a problem for another PR. :) 

Things are looking fantastic! Definitely up for handling this in a better way. Having a lot of dotfiles is slightly annoying

EyeCantCU:
One final thing - the Containerfile needs the service to be enabled:  `systemctl enable --global bluefin-dx-user-vscode.service`

castrojo (approved):

EyeCantCU (approved):
LGTM!

PR #752, opened 2023-12-23
Fix: Don't delete ganto-lxc4 copr since not everything is on image

m2Giles:
Don't delete the ganto-lxc4 repo. There are additional packages that are installable. Namely lxd-tools, incus-tools, and lxd-migrate. These tools can convert traditional lxc style containers to the managed format. lxd-migrate can convert a physical machine to a VM. The corresponding tool is in incus-tools.


m2Giles:
Actually. These missing binaries from the repo are pretty small. Instead of leaving the ability to layer them, just have them on the -dx image. Would make rebasing easier as well.

PR #753, opened 2023-12-23
fix(-dx): Don't remove ganto-lxc4 repo.

m2Giles:
Not all packages from ganto-lxc4 repo are installed on the image. Additionally, these missing packages conflict and can't be installed without changing the build script. lxd-tools and incus-tools are the packages in conflict.

This allows a user to layer the respective tools if necessary. Incus-tools provides tooling of both lxd-tools and lxd-migrate.


castrojo:
Standing by on this, we're investigating options packaging up incus so we can have all the binaries we want. Also seeing talking to upstream, seeing if we can find someone to just put this in distro where it should be.

m2Giles:
I'm going to close this. If reenabling the repo is in the best interest I can make a new pull request.

PR #757, opened 2023-12-25
feat: Fix caps-lock input delay

fiftydinar:
This is one of the 1st issues which ex-Windows users notice when they switch to Linux. I am one of those users who uses Caps-Lock instead of Shift for uppercase letters.

This will get rid of input delay, while the "light indicator" delay is still the same (I mentioned this distinction to not confuse you if you don't "see" any changes).

To test, type "Helo" repeatedly & quickly using Caps Lock.
You will no longer see "HElo" in your texts.

Fix took from here:
https://forum.manjaro.org/t/caps-lock-behaviour-wayland/79868/8

### DIFF
--- a/usr/etc/dconf/db/local.d/01-ublue
+++ b/usr/etc/dconf/db/local.d/01-ublue
@@ -23,6 +23,9 @@ picture-options='zoom'
 primary-color='000000'
 secondary-color='FFFFFF'
 
+[org/gnome/desktop/input-sources]
+xkb-options=['caps:ctrl_modifier']
+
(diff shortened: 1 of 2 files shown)


castrojo:
This appears to be X specific?

fiftydinar:
> This appears to be X specific? EDIT: What I mean to ask is, the Wayland session would remain unaffected?

No, it affects both sessions.

castrojo:
I was just curious, LGTM!

castrojo (approved):

PR #759, opened 2023-12-28
feat: incus enable script

bketelsen:
Adds interactive shell script to enable Incus on bluefin-dx


dx/usr/bin/bluenfin-incus
p5, line 1: Please rename to `bluefin-incus` unless there's another meaning I'm missing

dx/usr/bin/bluenfin-incus
```diff
@@ -0,0 +1,143 @@
...
+then
+    echo "You are already on a developer image"
+else
+    echo ""
+    echo "Before we can enable Incus on this system"
+    echo "the current system needs to be on a developer (dx)"
+    echo "image. Please run 'just devmode' to rebase to a developer image"
+    echo "and reboot your system when the rebase is finished."
+    exit
+fi
```
p5, line 45: I don't think you need most of this logic.  This script will only be present on the `dx` images, so there's no need to check if this is a DX image or not

p5 (commented):
Not a bash wizard, so would like someone else to check over this, but mostly LGTM.  

Added a few comments though on some stuff I am not too sure about.

dx/usr/bin/bluenfin-incus
p5: I guess there's no harm in leaving it here though.

dx/usr/bin/bluenfin-incus
bketelsen: d'oh!

dx/usr/bin/bluenfin-incus
bketelsen: i flip/flopped on this, thinking someone might see it in the repo, or in a github search. I can go either way. Tiebreaker @castrojo ?

dx/usr/bin/bluefin-incus
```diff
@@ -0,0 +1,154 @@
...
+if grep -q "/var/ublue-os/image" <<< $CURRENT_IMAGE
+then
+    echo ""
+    echo "Before we can enable Incus on this system"
+    echo "the current system needs an update. Please run 'just update'"
+    echo "and reboot your system when the update is finished."
+    exit
+fi
+
+if grep -q "bluefin-dx" <<< $CURRENT_IMAGE
```
EyeCantCU, line 25: I would only ship this script on DX so that this check isn't needed

dx/usr/bin/bluenfin-incus
p5: If you're thinking about if this is used outside ublue, maybe some checks to make sure the required CLI tools are installed, rather than checking the exact image?
I would expect this script to not function at all on a random Ubuntu VM since it's expecting the `ublue-os/image` file to be present, exiting if it is not.

E.G. 
```
CHECK INCUS_CLI EXISTS
IF NOT EXIST:
  OUTPUT: "Incus CLI is not installed.  Please install it.  If on Bluefin, run `just devmode on`"
ELSE:
  CONTINUE
```

just/custom.just
```diff
@@ -183,6 +183,11 @@ garden:
         echo ""
     fi
 
+# Install and configure Incus
```
EyeCantCU, line 186: The logic to determine whether or not a user needs to rebase to DX could be moved here so that shipping the script on DX only is inconsequential

EyeCantCU (dismissed):
This looks great as is but I've left a couple comments. I think it would be best to ship the script solely on DX and implement the logic for determining whether or not to rebase in the script added to the justfile instead of handling it in the main script

bketelsen:
@p5 @EyeCantCU Please take another look, comments resolved.

p5 (approved):
LGTM

EyeCantCU (approved):
Looks fantastic!

PR #762, opened 2023-12-30
feat: add dive command

castrojo:
https://github.com/wagoodman/dive

This is useful for inspecting container sizes, etc but needs to be on the image.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -139,7 +139,7 @@ RUN rpm-ostree override replace --experimental --from repo=copr:copr.fedorainfra
 RUN wget https://github.com/docker/compose/releases/latest/download/docker-compose-linux-x86_64 -O /tmp/docker-compose && \
     install -c -m 0755 /tmp/docker-compose /usr/bin
 
-COPY --from=cgr.dev/chainguard/dive:latest /usr/bin/kubectl /usr/bin/dive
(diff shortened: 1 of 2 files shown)


Containerfile
```diff
@@ -139,6 +139,7 @@ RUN rpm-ostree override replace --experimental --from repo=copr:copr.fedorainfra
 RUN wget https://github.com/docker/compose/releases/latest/download/docker-compose-linux-x86_64 -O /tmp/docker-compose && \
     install -c -m 0755 /tmp/docker-compose /usr/bin
 
+COPY --from=cgr.dev/chainguard/dive:latest /usr/bin/dive /usr/bin/dive
 COPY --from=cgr.dev/chainguard/flux:latest /usr/bin/flux /usr/bin/flux
 COPY --from=cgr.dev/chainguard/helm:latest /usr/bin/helm /usr/bin/helm
 COPY --from=cgr.dev/chainguard/ko:latest /usr/bin/ko /usr/bin/ko
```
p5, line 145: Since these are all single binaries, _maybe_ we should start fetching them direct from GitHub Releases so we don't need to verify the signatures?

p5 (approved):

Containerfile
castrojo: This part and the github action are kind of an SBOM demo I'm working on.

PR #764, opened 2023-12-31
fix(-dx): Workaround swtpm SELinux issues

m2Giles:
swtpm has the wrong SELinux labels. You This is a deficiency with the OCI build process for ostree.

Libvirt is shipped with -dx image. This pulls in swtpm. On the discord there has been numerous callouts to issues with swtpm being broken. This should be a sufficient workaround for the time being. Compared to my original implementation, I've changed the binary location from /usr/local/bin to /tmp and using a tmpfile to create the required directory for the swtpm-rootca. swtpm is only 42K and we do not set noexec on tmp.


m2Giles:
Instead of calling a script, we can do everything just in the oneshot. We can also clean up after ourselves with ExecStop lines. Instead of using a .mount file, I call mount directly in the unit file. This "works" but a mount file would be better but seemed to slow down boot for some odd reason and makes it so we can't as easily clean up after ourselves. Additionally use tmpfiles.d to create a copy in the beginning and make sure that /var/lib/swtpm-localca exists and is owned by tss.

Additionally, using /tmp doesn't seem to work consistently so placed back in /usr/local/bin. I prefixed the binary with a . to help avoid a name collision and clean it up when the service exits.

bketelsen (dismissed):
YOLO

bketelsen (approved):

KyleGospo (approved):

PR #767, opened 2023-12-31
fix: remove brew update-reset

castrojo:
We don't need this anymore.


m2Giles (commented):
This file isn't copied into the bluefin-cli Containerfile currently.

castrojo:
Yeah not sure if we need this file at all anymore?

m2Giles:
In my state maintain I rewrote it.

PR #770, opened 2024-01-01
fix: call the correct binary from the service unit

castrojo:


p5 (approved):

castrojo:
Unreleated build failure, merging.

PR #780, opened 2024-01-03
fix: remove `--disable-content-trust` push flag

p5:
I do not believe this flag is necessary.  If someone knows the origin of this, and we still need it, happy to close this PR.

> When transferring data among networked systems, trust is a central concern. In particular, when communicating over an untrusted medium such as the internet, it is critical to ensure the integrity and the publisher of all the data a system operates on. You use Docker Engine to push and pull images (data) to a public or private registry. Content trust gives you the ability to verify both the integrity and the publisher of all the data received from a registry over any channel.


castrojo:
I see we have this in the main repo and probably others as well. I don't recall why this is in there but looks like the toolboxes pushed just fine without it?

castrojo (approved):

PR #790, opened 2024-01-04
feat: Switch to system level flatpaks

KyleGospo:


EyeCantCU (dismissed):
Looks good. Consistency between Bluefin and Bazzite will be nice

usr/etc/flatpak/user/install
```diff
@@ -1,22 +0,0 @@
-org.mozilla.firefox
```
p5, line 1: Maybe a stupid question, but will this lose an existing users' configuration?  Like bookmarks, themes and whatever else you can store in some of these apps?

p5 (dismissed):

usr/etc/flatpak/user/install
KyleGospo: We definitely need to test this, but it should be kept since the system and user Firefox both use the same directory for their configuration and we don't specify the option to delete app data to the remove command.

castrojo:
Should we wait to see what the flatpak installation method will be in the installer?

ispanos:
@KyleGospo  Can you please add a new line at the end of the list? It would be easier to append more packages with `echo` when building on top of bluefin with startingpoint.

ispanos:
> Should we wait to see what the flatpak installation method will be in the installer?

Can we merge the fix and decided about the installation later?

KyleGospo:
@castrojo Rebased and ready for review

castrojo (approved):

PR #794, opened 2024-01-05
chore: clean-up dotfiles from home dir

ispanos:


m2Giles (approved):
This looks good. Matches what we are doing for system state in /etc.

PR #797, opened 2024-01-06
fix: switch back to distro gnome-c-c

### LINKED ISSUE #795: [Security][Bug] Missing use-after-free fix from GNOME 45.2 packages, causes crashes
### Describe the bug

run this on bluefin:

```
rpm -qa | grep gnome-control
```

Observe that the version is 45.1 and not 45.2

Also, open gnome-control-center on bluefin using a hardened memory allocator, and observe that the process crashes due to the memory allocator blocking the use-after-free vulnerability.

The version should be 45.2

_No response_

https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/2647
https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/2012
https://github.com/secureblue/secureblue/issues/152

castrojo:
The webapp stuff isn't going to work out for us so switching back to distro control center. Should fix #795


m2Giles (commented):
Does removing the patched G-C-C also remove the VRR options?

KyleGospo:
It does, web app patch is in gnome software

castrojo:
Oh, I'm digging in the wrong place, thanks!

PR #820, opened 2024-01-17
Add Terra repository to ublue-os bluefin image

td5038:
I've been working a little bit on adding Terra stuff to this image so that other apps that are very useful can be added.


td5038:
@castrojo

td5038:
ok

PR #826, opened 2024-01-20
feat: match bazzite's logomenu

castrojo:


KyleGospo:
You need xdg until you drop 38 or set up a way to only apply this to 39 builds

PR #836, opened 2024-01-26
feat: Switch to TuneD from PPD (F39+)

KyleGospo:


castrojo:
Holding on this one until I test it due to it affecting the -framework images

ispanos:
Is this change comming to [F40](https://fedoraproject.org/wiki/Changes/TunedReplacesPower-profiles-daemon#Tuned_Replaces_Power-profiles-daemon) or not? I don't understand.

> The power-profiles-daemon API should be ported to tuned to provide the function to the application that uses power-profiles-daemon API, such as gnome-shell and gnome-control-center.

- [wiki/Changes](https://fedoraproject.org/wiki/Changes/TunedReplacesPower-profiles-daemon#Feedback)

Is this implemented already? I have never used tuned and I don't know if the "Power Mode" in Gnome will be broken or not.

castrojo:
The GNOME power toggle will work, we were waiting for it to work before trying this. Fedora has decided not to switch this cycle but we're keen to see how this works out for people since it's much more powerful and allows for better end-user customization of their power curves.

Normally we don't like to diverge from Fedora for things like this but we've got time until F40 and power management is always a pain point so we're going to see if we can get better results.

castrojo (approved):

castrojo (dismissed):

castrojo (approved):

PR #857, opened 2024-01-29
fix: improve ujust bluefin-cli

m2Giles:
Enables the use of all of toolboxes in ublue-os/toolboxes
Dynamically fetches the Quadlets and makes systemd targets
Moves the inline ujust command to it's own script


HikariKnight:
I will have to head to bed (it is midnight here), but i will continue looking at this tomorrow unless someone else picks up where i left off.
This is all i found so far and the lack of `.service` unit files for the quadlets as i mentioned in my question to you on discord since i have not worked with quadlets, but it seemed like the script expects them so i figured i would ask.

HikariKnight:
Would also recommend adding an exit in case the user cancels a selection (for example when selecting a container they can press ESC to cancel, this currently seems to lock the script up)

m2Giles:
That seems smart. Each question should resolve to a variable. If a variable is empty bail out. Right now a SIGINT or empty string will put the script in a bad state.

I also should move all questions to occur before any execution. For the loops that might be an issue but `trap` would probably be sufficient to have a clean exit for those.

PR #858, opened 2024-01-29
feat: Mostly switch from dconf to gschema override

fiftydinar:
This was something that Arcitec talked about back in May last year.

Advantages:
- Troubleshooting logs support during build-time.
You get logs if some schema is depreciated, when some key is no longer valid & such.
- You don't mess with dconf database, gsettings takes your override changes directly instead as a more clean way

The reason why it's named `zz` in the beginning is to ensure that nothing will override Bluefin gschema. Alphabetic gschemas take advantage over numeric ones:

- 00-99 (99 takes the advantage)
- a-z (z takes the advantage)

Format of schema & keys is exactly like gsettings.

All of those should be available & work straight away when "converted" from dconf. However, it can occur rarely that some RPMs don't place their gschema to `/usr/share/glib-2.0/schemas` location. Example of this is Caffeine Gnome extension in RPM. They should fix that in spec file. Dconf doesn't care about gschema location, so it sees everything.

Only thing that gschema override doesn't support is setting locks & relocatable schemas, which are used for folders. That's why I did not remove them.

Please review & test this thoroughly if every setting is applying properly before approving.


bsherman:
For the record, I'm interested in testing this out as it could be a good improvement to how we (bluefin, bazzite, and friends) make these types of customization.

But it's not top on my list, so I'd like to leave it open and come back to it.

bsherman:
I did some more testing on this... i wonder if we don't want to use the prefix `za-`  or `z0-` so that it still sorts very near the last and it's unlikely anything would override it, but, it could still be overridden if a downstream user intended to do so.

fiftydinar:
> I did some more testing on this... i wonder if we don't want to use the prefix `za-` or `z0-` so that it still sorts very near the last and it's unlikely anything would override it, but, it could still be overridden if a downstream user intended to do so.

Yeah, we can do that, good idea

bsherman:
I think this looks pretty good. I've used the information in this PR to update my own ( https://github.com/bsherman/ublue-custom/pull/6 ).

@castrojo you should be the approver on this. It should make gsetting overrides more reliable/less prone to error, by failing when one is incorrect. It does add a small bit of complexity in the need for a few things (like custom keyboard shortcuts and app folders) to be have modifications in both gsettings override and in dconf db.

I'd approve it. :-)

fiftydinar:
In case someone did not read commit message:

I tested my PR with separate image & with latest fixes, everything works, except 1 thing.

Logo-Menu extension settings don't work.

There are 2 ways to fix this.

The best way to fix this is to modify RPM spec-file to copy its gschema.xml to `/usr/share/glib-2.0/schemas` instead to it's own extension folder.

2nd worse way is to move those settings to dconf.

I think that Kyle is handling LogoMenu fork, so I think he can easily fix it.

To test my changes, you can rebase to my image with this command:

`rpm-ostree rebase ostree-unverified-registry:ghcr.io/fiftydinar/bluefin-gschema-test:39`

bsherman:
> I tested my PR with separate image & with latest fixes, everything works, except 1 thing.
> 
> Logo-Menu extension settings don't work.
> 
> There are 2 ways to fix this.
> 
> The best way to fix this is to modify RPM spec-file to copy its gschema.xml to `/usr/share/glib-2.0/schemas` instead to it's own extension folder.

@KyleGospo what do you think about this modification to Logo-Menu to support schema override?

fiftydinar:
@bsherman Since KyleGospo is busy with Bazzite stuff, shall we move Logo-Menu settings to dconf & merge this?

fiftydinar:
I will close this PR, as there is no interest in this at the moment.
Structure also changed radically in some areas, so If I wanted to do this, I would need to begin from scratch.

PR #859, opened 2024-01-29
fix: set Prompt theme to Catpuccin

castrojo:
If we set it to just Catpuccin it works in both light and dark.


castrojo:
Further testing shows this doesn't do what we want. Will revisit later.

PR #870, opened 2024-02-02
chore: match bazzites topgrade.toml

m2Giles:
Have include field.


castrojo (approved):

KyleGospo:
@castrojo need to revert this, unless that file exists in /etc this breaks topgrade.

PR #886, opened 2024-02-10
feat: add ability to set Prompt terminal transparency with ujust

dperson:


castrojo:
@KyleGospo I think you'll dig this!

castrojo (approved):
Thanks!

PR #889, opened 2024-02-11
feat: improve prompt integration with quadlets

m2Giles:
Prevent multiple profiles from being created for same container
Have list of known containers
Set palettes for known containers.
Ability to set default profile and palettes when creating profiles
Handle opacity if already set on default profile


castrojo (approved):
Forward unto dawn!

Issue #891, opened 2024-02-11
Make tailscale optional


dylanmtaylor:
Can we have tailscale as an option during inital setup but not enabled out of the box? I think minimal by default is better. The toggle can enable the extension if users want it.

dylanmtaylor:
To be fair, I'm 100% fine w/ tailscale being installed. I just don't like the extension being enabled which adds something to the quick settings dropdown since I don't use it.

PR #896, opened 2024-02-12
feat: remove tailscale extension from pre-installed packages (#891)

dylanmtaylor:
This could be installed by users later via GNOME extensions but feels really out of place in the default installation.


castrojo:
Tailscale is considered a major feature for Bluefin! Sorry this is the intended experience.  Here are the docs for that if you want to turn it off: https://universal-blue.discourse.group/docs?topic=166

PR #907, opened 2024-02-14
feat(ci): add cosign keyless verification via github OIDC

bayou-brogrammer:
This PR adds the keyless verification from [cosign docs](https://github.com/sigstore/docs/blob/main/content/en/signing/overview.md).

cosign keyless requires two inputs:
- certificate-oidc-issuer
- certificate-identity-regexp

For github, our template follows:
- **certificate-oidc-issuer** => `https://token.actions.githubusercontent.com`
- **certificate-identity-regexp** => `https://github.com/${{ github.repository }}`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,5 @@
 name: Build and Push Image
+
 on:
   schedule:
     - cron: "40 16 * * *" # 16:40 UTC everyday
@@ -16,6 +17,7 @@ on:
     paths-ignore:
       - "**.md"
   workflow_dispatch:
+
 env:
   IMAGE_NAME:
   AKMODS_FLAVOR:
@@ -38,9 +40,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image_flavor:
-          [main, nvidia, asus, asus-nvidia, framework, surface, surface-nvidia]
         base_name: [bluefin, bluefin-dx]
+        image_flavor: [main, nvidia, asus, asus-nvidia, framework, surface, surface-nvidia]
         major_version: [38, 39]
         include:
           - major_version: 38
@@ -51,9 +52,6 @@ jobs:
             is_latest_version: true
             is_stable_version: true
             is_gts_version: false
-    outputs:
-      tags: ${{ steps.meta.outputs.tags }}
-
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,40 +1,45 @@
 name: Build and Push Image
 on:
   schedule:
-    - cron: '40 16 * * *'  # 16:40 UTC everyday
+    - cron: "40 16 * * *" # 16:40 UTC everyday
   merge_group:
   pull_request:
     branches:
       - main
       - testing
     paths-ignore:
-      - '**.md'
+      - "**.md"
   push:
     branches:
       - main
       - testing
     paths-ignore:
-      - '**.md'
+      - "**.md"
   workflow_dispatch:
 env:
-    IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
+  IMAGE_NAME:
+  AKMODS_FLAVOR:
+  IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -234,18 +241,50 @@ jobs:
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: sigstore/cosign-installer@v3
+
+      - name: Sign the images
+        run: cosign sign -y ${TAGS}
+        env:
+          TAGS: ${{needs.build_and_push.outputs.tags}}
```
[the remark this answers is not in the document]
p5: I feel the easiest way to do this would be to move the sign step back to the build jobs, otherwise we'd need to have some logic to extract matrix information from the job metadata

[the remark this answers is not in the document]
bayou-brogrammer: The easiest solution would be to move it back into the build job. 

If we wanted to separate the jobs, it would require a few extra steps:

https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://github.com/orgs/community/discussions/26284&ved=2ahUKEwjiiaukrauEAxVzkGoFHXGPA4AQFnoECBsQAQ&usg=AOvVaw3iDi-nAVM0p_-IJrisi69Z

.github/workflows/build.yml
```diff
@@ -218,32 +223,39 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      # Sign container
+      # Sign container via github OIDC
       - uses: sigstore/cosign-installer@v3.4.0
         if: github.event_name != 'pull_request'
 
-      - name: Sign container image
+      - name: Sign the images
```
p5, line 230: Without the public and private keys, how will we set the OS configuration to allow images from this repo?

p5: Right now, we have configured the ublue certs as trusted in all images, but now we're not using those certs, how will it work?

bayou-brogrammer: So i might be a bit out the loop on the entire process, but I thought we could now use our github to validate via cosign

```
cosign verify --certificate-oidc-issuer "https://token.actions.githubusercontent.com" --certificate-identity-regexp https://github.com/ublue-os/bluefin <TAG>
```

which will allow for the image to be verified against our OIDC signature

gerblesh: recently got my bluebuild image to sign with OIDC and the problem stems from that now the image is no longer signed with the same key, yet the `policy.json` still expects it to, so updating fails:
```
user@localhost ~ [1]> rpm-ostree update
Pulling manifest: ostree-image-signed:docker://ghcr.io/gerblesh/uhypr:latest
error: Creating importer: Failed to invoke skopeo proxy method OpenImage: remote error: cryptographic signature verification failed: invalid signature when validating ASN.1 encoded signature
```
we could try creating special tags and systemd units for rebasing/fixing the `policy.json` but I don't know what's the best way forward for this, and it would probably be ideal to keep workarounds like the ones I just mentioned to a minimum

bsherman: I mentioned this in Discord, but appropriate to ask here... 

What benefit(s) does signing with OIDC provide bluefin (or any ublue-os image)?

Unless I'm missing something, to trust the OIDC signed image, the policy,json still needs to provide some information to know how to validate the signatures... so one still can't do a direct rebase from upstream Silverblue to `ostree-image-signed:docker://ghcr.io/example-org/oidc-signed-image:latest` .

And it seems your example here where skopeo fails demonstrates this problem.

gerblesh: removing the configuration for the individual image allows a rebase. Signing with OIDC keyless means the signature changes every time and is verfied using github's provider. The default policy.json will work on rebase if you don't override the signature verification with pubkey auth

gerblesh: cosign's docs for keyless signing are here: https://docs.sigstore.dev/signing/overview/

JasonN3: https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md#sigstoresigned explains how to specify OIDC verification in policy.json. There's also an option for `subjectEmail`, which will ensure the image is signed by the right account and not a random GitHub account.

p5:
Converted to draft to avoid accidental approval/merge before we have figured out how it will work.
I do not believe our current setup is broken, so there's no rush to switch over to OIDC at this moment.

PR #912, opened 2024-02-16
fix(fish): address overriden config

bayou-brogrammer:
This seemed like the easier route which is to move our fish functions into etc so they are not overridden when fish installs.


m2Giles:
There appears to be no changes at all now?

bayou-brogrammer:
> There appears to be no changes at all now?

its a file change from `/usr` to `/etc/`

m2Giles:
There is a vendor directory as well. I was using that location for bluefin-cli.

My only concerns with /etc is that we lose out on image changes if they make any local changes in /etc

m2Giles (approved):

PR #917, opened 2024-02-19
chore: Remove unneeded tuned package

KyleGospo:


bsherman (approved):
Good to remove this since it pulls gcc, make, kernel devel packages, etc.

PR #925, opened 2024-02-23
fix: remove nix shortcuts

castrojo:
Matching the removal in config. We'll reinvestigate readding devbox when nix doesn't break selinux.


bsherman (approved):
:+1:

PR #937, opened 2024-02-24
chore(ci): combine ISO and release process into one

castrojo:
This should build ISOs first, then generate a changelog.


bsherman (approved):
Looks good.

PR #948, opened 2024-02-26
feat: bring back the brew

bketelsen:
This PR adds `gcc` to the bluefin package list (to be inherited by -dx as well) and adds a script in /etc/profile.d/ to source the brew shellenv script only from interactive bash terminals.

It DOES NOT address zsh or fish. TODO

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -7,6 +7,7 @@
 				"evtest",
 				"fish",
 				"firewall-config",
+				"gcc",
 				"glow",
 				"gum",
 				"gnome-shell-extension-appindicator",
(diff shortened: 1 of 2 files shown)


castrojo:
Ok what do I need to do to my existing system to get this working?

castrojo (approved):

bketelsen:
> Ok what do I need to do to my existing system to get this working?

after rebasing to this build, install homebrew the normal way. BUT... don't follow the instructions at the end to `eval linuxbrew shellenv`, and don't modify your .bashrc.  The script in /etc/profile.d/brew.sh does that for you.

PR #951, opened 2024-02-26
feat: create open -> xdg-open shell alias

p5:
Adds an alias for the MacOS folks so that `open <file>` opens that file in the default application.  It means they don't need to know about xdg-open and have the massive terminal output from that command.


castrojo (approved):

castrojo:
If the rest pass except the one image please force merge. 🗡️

bketelsen:
nice!

PR #952, opened 2024-02-27
fix: Allow brew-install to escalate when needed, fixes permission issue.

KyleGospo:
Also renames the flatpak installer to instead be a "networked tasks manager", of which Brew joins the roster.

### DIFF
--- a/usr/libexec/ublue-user-networked-tasks
+++ b/usr/libexec/ublue-user-networked-tasks
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 # Script Version
-VER=3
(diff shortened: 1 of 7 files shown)


usr/libexec/ublue-user-networked-tasks
```diff
@@ -50,4 +50,11 @@ fi
 
 notify-send "Flatpak installer" "Finished installing user flatpaks" --app-name="Flatpak installer" -u NORMAL
 
+# Install brew
+if [[ ! -x /var/home/linuxbrew/.linuxbrew/bin/brew ]]; then
+  NONINTERACTIVE=1 /usr/libexec/brew-install
```
castrojo, line 55: Not CI=1?

castrojo (approved):

Issue #365, opened 2023-08-03
pmlogger has selinux & directory issues


bketelsen:
will leave logs when I'm back at my desk
pmlogger service fails with selinux problems

castrojo:
I think this is the same issue: https://github.com/ublue-os/ucore/issues/78

(We also removed pmlogger because of this)

PR #953, opened 2024-02-27
fix: add pmcd log directory with tmpfiles.d

bketelsen:
noticed pmcd log directory was missing and service was failed.  Here's a tmpfiles.d fix for it.


bsherman:
Relates: #365

PR #961, opened 2024-02-28
fix: clarify instructions

castrojo:
Just in case


bsherman (approved):
Obviously :-)

castrojo:
I genuinely stumped myself when I ran it in testing because it was like "don't do what this tells you to do" and then the first thing it does is ask for my sudo password lol.

PR #975, opened 2024-02-29
fix: Make this check if brew is installed before trying to run it.

dylanmtaylor:


bketelsen (approved):

bketelsen:
I approved this and it's broken.

PR #980, opened 2024-02-29
fix: check for directory, not command

bketelsen:
check the directory, not the command for sourcing brew shellenv


p5 (approved):

dylanmtaylor:
I edited 3 files in my PR. You probably need to update the other two as well.

Issue #983, opened 2024-02-29
/etc/profile.d/brew.sh won't overwrite on update


bketelsen:
### Describe the bug

brew.sh doesn't update

### What did you expect to happen?

it should

### Extra information or context

we need to add it to the ujust brew task to replace /etc/profile.d/brew.sh

bronson:
Probably also add a warning inside the file: ### DO NOT EDIT, WILL BE REPLACED

p5:
Would it work if we moved the script to /usr and symlinked the script to profile.d?

bketelsen:
good question

bronson:
Huh! Symlinking from /usr seems like it solves everything: 

- automatically updates
- doesn't silently throw away user changes
- file contents are immutable
- user can override if they really want

One question: what do you do when it's time to delete the file? Easiest thing would be to empty the file (0 lengnth) and leave the link in /etc untouched. But that's crufty. Probably still need a script to clean up.

Other than that, this seems like a solid idea.

m2Giles:
Symlink would be the best bet.

I currently extend the brew stuff in my home directory to make sure bash completion is working. Need to get that stuff added to brew.sh as well.

m2Giles:
Seeing the just command do a force overwrite of the file. We should do a symlink.

If there is a symlink it's still managed by system, if a file we can prompt to remove for the symlink or have it remain in place.

Symlink means we can update the file and not worry if the file in /etc is getting properly updated.

I think a lot of our customizations could move to this method.

PR #984, opened 2024-02-29
fix: replace profile brew.sh on just brew command

bketelsen:
this should overwrite /etc/profile.d/brew.sh with the one in /usr/etc regardless of whether the brew install actually happens in the test above.


bketelsen:
fixes #983

PR #985, opened 2024-02-29
fix: Increase Nofile limit to fix installation of some brew application

fiftydinar:
This follows recommendation from Lutris for making Wine Esync working. This one should be more than enough for brew programs.

https://github.com/lutris/docs/blob/master/HowToEsync.md

I increased soft limits too, just in case.


bketelsen (approved):
LGTM

Issue #670, opened 2023-11-19
Increase open files


castrojo:
The open file limit is still set too low. `brew install texlive` failed due to lack of open files. 

`fs.file-max = 500000` in `/etc/sysctl.conf` fixes it. We should consider bumping up the default.

PR #988, opened 2024-03-01
fix: correct ulimit nofile issue with brew

bsherman:
`brew install texlive` was still failing with:
`Error: Too many open files @ rb_sysopen`

This sets a functional limits.conf soft nofile limit of 4096, but also sets the same default in systemd/user.conf which is required for the ulimit value to be respected in terminals running under Wayland/Xorg.

Relates: #670

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -114,6 +114,7 @@ RUN wget https://copr.fedorainfracloud.org/coprs/ublue-os/staging/repo/fedora-"$
     echo "Hidden=true" >> /usr/share/applications/nvtop.desktop && \
     echo "Hidden=true" >> /usr/share/applications/gnome-system-monitor.desktop && \
     rm -f /etc/yum.repos.d/_copr_che-nerd-fonts-"${FEDORA_MAJOR_VERSION}".repo && \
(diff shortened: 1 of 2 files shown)


noelmiller (approved):

KyleGospo (approved):

KyleGospo:
🛌

fiftydinar:
Is my fix wrong?

KyleGospo:
> Is my fix wrong?

Soft limit gets overridden by systemd, so your fix only applies to the hard limit which this isn't affected by.

bsherman:
> Is my fix wrong?

I did some testing because there were still reports of the error in PR description/commit message.

I didn't realize there had already been commits to address it, but no, sadly, your fix had not resolved the issue.

Your fix worked for a user logged in at a console tty or via SSH :grinning:, but not for users using Bluefin's default bash shell via Ptyxis or gnome-terminal under gnome.


I discovered that in this scenario, systemd is playing a primary role setting `nofile` rather than pam_limits.  There is some interplay with the two. For example, I could get pam_limits to raise or lower the `hard nofile` limit so that it differs from systemd's default. However, I could NOT get pam_limits to raise the `soft nofile` limit, only reduce it.

In addition, it seems slightly dangerous to raise the soft limit to equal the hard limit, so I did remove those lines in the brew-limits.conf.

PR #1003, opened 2024-03-04
chore(just): cleanup and standardize just recipes

HikariKnight:
Drafting this so we can start a discussion.
Unlike the just reorg done in https://github.com/ublue-os/bazzite/pull/765
Bluefin has way less just recipes so I would like to prod and hear what the reorg needs will be and if the just files should be split up in groups like we did in bazzite.

In bazzite we focused on merging together recipes that involved the same thing under the `configure-` and `setup-` verbs.
- `configure` for things that already exist on the image
- `setup` for anything that would require installation before configuring. (we also have `install-foo` for anything that just installs without any configuration or removal steps)

shortcuts like `ujust benchmark` would stay as they are however i can merge all the `ujust zsh|fish|bash` into `ujust configure-shell` with an alias named `ujust shell`

Other than this there does not seem to be anything else to do other than just some minor cleanup and change recipes to follow the standard we decided on in https://github.com/ublue-os/config/pull/181

### DIFF
--- a/just/custom.just
+++ b/just/custom.just
@@ -156,14 +156,14 @@ jetbrains-toolbox:
 # Install and configure Incus
 incus:
     #!/usr/bin/env bash
+    source /usr/lib/ujust/ujust.sh
     CURRENT_IMAGE=$(rpm-ostree status -b --json | jq -r '.deployments[0]."container-image-reference"')
-
     if grep -q "bluefin-dx" <<< $CURRENT_IMAGE
     then
         echo 'Installing and configuring Incus.'
         /usr/bin/bluefin-incus
     else
-        echo "Developer mode is currently ${CURRENT_STATE}."
+        echo "Developer mode is currently ${b}${red}Disabled${n}."
         echo "Run `just devmode` to turn on Developer mode."
         exit
     fi
--- a/just/custom.just
+++ b/just/custom.just
@@ -164,7 +164,7 @@ incus:
         /usr/bin/bluefin-incus
     else
         echo "Developer mode is currently ${b}${red}Disabled${n}."
-        echo "Run `just devmode` to turn on Developer mode."
+        echo "Run \"just devmode\" to turn on Developer mode."
         exit
     fi
 


m2Giles:
Shells:
I think we have multiple just commands for switching default using various methods. Maybe make this one command and have a selector between them?

Configuration:
We don't have a lot of common verbiage. Most seem to be one direction setups. For one direction I think setup would be good verbiage. Toggles could be configure.

Overall we probably have a lot of things that could be removed. We also should consider not in lining so much stuff.

castrojo:
> Maybe make this one command and have a selector between them?

This would be great for shells!

HikariKnight:
Currently since there have not been any verbs before on anything, i have left in aliases to the old recipe names so that you all can remove them when the time is right for it

just/bluefin-system.just
```diff
@@ -0,0 +1,155 @@
...
+      echo "${USER}'s shell is currently ${CURRENT_SHELL}"
+      OPTION=$(Choose "fish" "zsh" "bash")
+    fi
+    if [[ "$OPTION" =~ zsh|fish ]]; then
+      sudo usermod $USER --shell /usr/bin/$OPTION 
+      printf "${USER}'s shell is now %s.\n" "$(cat /etc/passwd | grep ":$UID:" | cut '-d:' '-f7')"
+    elif [ "$OPTION" == "bash" ]; then
+      sudo usermod $USER --shell /bin/$OPTION 
+      printf "${USER}'s shell is now %s.\n" "$(cat /etc/passwd | grep ":$UID:" | cut '-d:' '-f7')"
+    fi
```
m2Giles, line 96: Why are we doing an if - elif here? Also how is this handling a cancel out during the shell selection? Option becomes empty string then?

just/bluefin-apps.just
```diff
@@ -0,0 +1,156 @@
...
+        else
+          echo "Capitalization matters when you type \"YES I UNDERSTAND\""
+        fi
+    fi
+    # if /etc/profile.d/brew.sh already exists, replace it with /usr/etc/profile.d/brew.sh
+    if [ -f /etc/profile.d/brew.sh ]; then
+        if [ -f /usr/etc/profile.d/brew.sh ]; then
+            sudo cp /usr/etc/profile.d/brew.sh /etc/profile.d/brew.sh
+        fi
+    fi
```
m2Giles, line 37: I get that we may want to do this, but instead of copying over, maybe a symlink might make more sense? This might be a change we need to do on our /etc/profile.d/brew.sh

If that's a file then the user has modified, if it's a symlink we won't have to update? Since this prompt doesn't go away, I would want to avoid overwriting user changes.

m2Giles (commented):
Left a few comments. Overall looks solid

just/bluefin-system.just
HikariKnight: If the string is empty it will just exit as it does not match, as there is no code coming after it.
Also i saw by default the shell is `/bin/bash` and not `/usr/bin/bash`
I know they are the same path due to symlinks but i did it more for my OCD and can just put them all together to `/usr/bin/$OPTION` :sweat_smile:

just/bluefin-apps.just
HikariKnight: Can change that, i think @bketelsen wrote that originally, might be a reason why it is not a symlink instead.
Will wait for them to reply before making that change though, but i am pretty sure a symlink will be fine.

bketelsen: no reason, symlink is a great idea. please proceed!

m2Giles (approved):
We'll wrap up the brew symlink in another PR. LGTM.

PR #1019, opened 2024-03-11
chore(ci): switch to upstream for ISO generation

noelmiller:
# Purpose

Jason has put a ton of work into his iso generation project we use upstream. I would like to switch to using it so we don't need to maintain a fork moving forward. I will add flatpak support in a separate PR.

### DIFF
--- a/.github/workflows/build_iso.yml
+++ b/.github/workflows/build_iso.yml
@@ -33,6 +33,7 @@ jobs:
           - major_version: 39
             image_tag: gts
     steps:
+      # Build ISO using Matrix
       - name: Build ISOs
         uses: jasonn3/build-container-installer@main
         id: build
--- a/.github/workflows/build_iso.yml
+++ b/.github/workflows/build_iso.yml
@@ -33,7 +33,6 @@ jobs:
           - major_version: 39
             image_tag: gts
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_iso.yml
```diff
@@ -39,24 +39,27 @@ jobs:
             image_tag: gts
     steps:
       - name: Build ISOs
-        uses: ublue-os/isogenerator@1.0.9
+        uses: jasonn3/build-container-installer@main
```
castrojo, line 42: I'd like to pin a version here instead of main. (We'll get our updates via dependabot)

castrojo (changes requested):

.github/workflows/build_iso.yml
noelmiller: main is being used for testing for now until he publishes a release.

castrojo: Gotcha!

castrojo (dismissed):

noelmiller:
For 38 GTS images, bootc switch command is not working. This is likely due to version of bootc that is shipped with gts. 
image

noelmiller:
Closing PR, moving to testing

PR #1022, opened 2024-03-12
chore(ci): Add system flatpaks

noelmiller:
# Purpose

Add system level flatpaks as part of ISO creation and combine workflows


.github/workflows/build_iso.yml
```diff
@@ -47,6 +47,7 @@ jobs:
...
           enable_cache_skopeo: "false"
+          flatpak_remote_refs: "runtime/org.kde.Platform/x86_64/6.5 runtime/org.freedesktop.Platform/x86_64/23.08 runtime/org.gnome.Platform/x86_64/45 runtime/org.gnome.Platform/x86_64/44 app/org.gnome.Calculator/x86_64/stable app/org.gnome.Calendar/x86_64/stable app/org.gnome.Characters/x86_64/stable app/org.gnome.Connections/x86_64/stable app/org.gnome.Contacts/x86_64/stable app/org.gnome.Evince/x86_64/stable app/org.gnome.Loupe/x86_64/stable app/org.gnome.Logs/x86_64/stable app/org.gnome.Maps/x86_64/stable app/org.gnome.NautilusPreviewer/x86_64/stable app/org.gnome.TextEditor/x86_64/stable app/org.gnome.Weather/x86_64/stable app/org.gnome.baobab/x86_64/stable app/org.gnome.clocks/x86_64/stable app/org.gnome.font-viewer/x86_64/stable app/com.mattjakeman.ExtensionManager/x86_64/stable app/com.github.tchx84.Flatseal/x86_64/stable app/io.github.dvlv.boxbuddyrs/x86_64/stable app/io.github.flattool.Warehouse/x86_64/stable app/org.fedoraproject.MediaWriter/x86_64/stable app/io.missioncenter.MissionCenter/x86_64/stable app/io.github.celluloid_player.Celluloid/x86_64/stable app/org.mozilla.firefox/x86_64/stable"
```
p5, line 50: Does it work if you use 

```
  flatpak_remote_refs: |
    runtime/org.kde.Platform/x86_64/6.5
    runtime/org.freedesktop.Platform/x86_64/23.08
    etc
```

It would make it easier to maintain, however I'm not sure how GitHub Actions reads that sort of input

noelmiller: Jason is already working on a better solution: https://github.com/JasonN3/build-container-installer/pull/66

We will be able to define each flatpak in a file with all their dependencies

p5: Awesome!  I had a quick look through the Flatpak docs hoping there was a way to extract the dependencies from the apps, allowing us to automatically generate the list from the primary apps, but it didn't look like there was.

noelmiller: I appreciate you looking into that, but yeah, it's very manual which... is not super fun..

fiftydinar: Can't you just define the app name in the flatpak list, like when using `flatpak install`?
Why is there a need to manually define dependencies? Those are very dynamically changing, so they add some maintainance burden over time.

flatpaks/mozilla_firefox
JasonN3, line 1: This installs Thunderbird and the thunderbird file install Firefox

noelmiller:
Closing PR due to rewrite

PR #1030, opened 2024-03-15
feat: add bootc

castrojo:
We're gonna need it anyway


noelmiller:
I believe bootc is already available in Bluefin? Last I checked it was there.

castrojo:
Oh derp, this is in main.

PR #1038, opened 2024-03-20
chore(ci): Use flatpak folders and dependency resolution [ISO]

noelmiller:
**TODO:** 

BLOCKER: https://github.com/JasonN3/build-container-installer/pull/86
- [x] test installing ISO images that were created.
- [x] Remove commented out lines for building and pushing signed images. Need to open new issue and separate PR for this.
- [x] pin build ISO action to specific version (will do this as part of merge into main.)
- [x] Merge into testing branch


.github/workflows/build.yml
```diff
@@ -251,3 +251,10 @@ jobs:
       - name: Exit
         shell: bash
         run: exit 0
+
+  build_iso:
+    name: build iso
+    needs: [check]
+    if: github.event_name == 'pull_request' || github.ref_name == 'testing'
```
bsherman, line 258: A few thoughts on this condition:
1. using OR here means all pull requests will try to build an ISO, maybe fine now, but I was originally thinking of `pull_request AND 'testing'`, though I'm not sure it would work
2. i think you may also want       `  if: needs.push-ghcr.result == 'success'` or at least to add that as another `&&`  so the container build won't try to build an ISO unless it was a clean success

.github/workflows/build_iso.yml
```diff
@@ -62,7 +103,8 @@ jobs:
           overwrite: true
 
       - name: Upload ISOs and Checksum to R2
-        if: github.event_name == 'workflow_dispatch'
+        #if: startsWith(github.ref, 'refs/tags/v')
+        if: github.event_name != 'pull_request'
```
bsherman, line 107: i think there could be some improvements related to this. As a reusable workflow, you can take inputs, so the input when a callable could set a flag to do this or not. It's a bit awkward at the moment since this file directly run and ALSO reusable, so i'm not sure who we'd get the desireable behavior in a clean way.

This is an example of why i'd suggested making this a reusable workflow that's never run directly. 
Then, when calling from the `build`  workflow, you would pass in `upload_iso: false` but when calling from `build_iso` you could pass `upload_iso: true`.

Also, the logic of determining the "PR specific tag" could go away, because an `override_tag` could be passed in when calling from `build` workflow, but if no override, could just use the `gts` or `latest` tag logic.

For the moment, I do think this works fine, just pointing out where it could be a bit cleaner.

.github/workflows/build.yml
noelmiller: Building an ISO on all pull requests would be ideal in my opinion. It means every single PR has to be able to successfully create a container and ISO in order to be merged. I don't know if an `AND` would work here.

.github/workflows/build_iso.yml
```diff
@@ -41,12 +81,13 @@ jobs:
           image_repo: ghcr.io/ublue-os
           variant: 'Silverblue'
           version: ${{ matrix.major_version }}
-          image_tag: ${{ matrix.image_tag }}
+          image_tag: ${{ github.event.number && format('pr-{0}-{1}', github.event.number, matrix.major_version) }} || ${{ steps.generate-tag.outputs.tag }}
```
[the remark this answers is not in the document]
noelmiller: Yeah, I thought about this. I could probably add an if statement at the end that if it's a PR, override everything else.

.github/workflows/build_iso.yml
noelmiller: Kyle specifically said he never wants to build a container and upload in the same action. Not allowing the workflow to be called directly would not allow Kyle to control when uploads happen.

.github/workflows/build.yml
bsherman: The vast majority of changes have no bearing on ISO building success. And its a long, slow process which eats our builder time. So every time Jorge switches a wallpaper or adds a font, we'd have to build ISOs. I personally don't see the need, which probably gets back to some of the discussion we had earlier. :smile: 

Regardless, even my suggestion of `PR && testing` may not help my argument much if changes are typically made to the testing branch.

.github/workflows/build_iso.yml
bsherman: I don't think I explained it well.

My original suggestion was to have a `reusable-build_iso.yml` to which is most of the contents here are moved.

Then `reusable-build_iso.yml` would never run standalone but instead would be called by:
1. `build.yml` pretty much how you have it in this PR
2. `build_iso.yml` a very minimal file similar to what i showed you in ucore, which handles the "normal" ISO builds

The `reusable-build_iso.yml` would define inputs, some examples:
1. `override_tag: string` which could default to empty string or be passed a value
2. `upload_iso: boolean` which must be given a value by the caller

Then `build.yml` would pass `upload_iso: false` and `override_tag: pr-21` (for example), while `build_iso.yml` would pass `upload_iso: true` and no tag override.

.github/workflows/build.yml
noelmiller: Yeah. Maybe we can have a switch or something in the workflow to turn on building ISOs if that is something needed during testing?

PR #1044, opened 2024-03-25
fix: disable topgrade git extension

castrojo:
We don't want to touch user git directories. Also alphabetize the list for sanity.


bsherman (approved):
Added `git` and did a little alphabetizing... nice...

p5 (approved):

castrojo:
From Kyle: Git only actions if you manually configure a repo path
Otherwise its implicitly disabled

Reverting for now.

PR #1064, opened 2024-03-29
chore(ci): enable 40 builds

castrojo:


castrojo:
Ok this is ready but we'll need to special case the GE copr for nvidia explicit sync to be 39 only.

bketelsen (approved):
SHIP IT

PR #1065, opened 2024-03-29
fix: Add user flathub repo if --user flatpaks are found

noelmiller:
User reported in Discord that they were unable to update their flatpak apps after the recent changes to the user flatpak service. This change seeks to add the --user flathub repo only if user flatpaks are found.

**BLOCKER:**
I need additional testing with the service file which requires me to set up a local registry to rebase from. When doing manual testing in a VM, it will add it to the users home directory, not sure what the service will do, I want to test that.

### DIFF
--- a/usr/libexec/ublue-flatpak-manager
+++ b/usr/libexec/ublue-flatpak-manager
@@ -1,7 +1,7 @@
 #!/usr/bin/bash
 
 # Script Version
-VER=1
+VER=2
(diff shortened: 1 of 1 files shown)


p5 (approved):
LGTM

castrojo:
I tested in a VM and the user remote survived an upgrade just fine. Same thing with my laptop. We may not need this, let's see if we get more feedback.

castrojo:
We ended up not needing this, thanks tho.

PR #1072, opened 2024-03-30
feat: Add a ujust action to install ollama quadlet

anoldguy:
Running `ujust ollama` should:

1. Interrogate the machine's hardware, looking for nvidia (`which nvidia-smi` under the assumption that if you have the drivers installed, you *likely* have nvidia hardware...) and amd (`lspci | grep ' VGA ' | grep -sq Radeon`).
2. If multiple are found, present the user with a choice of `Nvidia (CUDA)` or `AMD (ROCm)`.
3. Create a quadlet at `~/.config/containers/systemd/ollama.container` with the appropriate bits set for the choice detected or selected by the user.
4. Prompt the user to install the `ollama` cli via `brew install ollama`.

I'm unsure about 2 things, and would appreciate some feedback.

1. Templating the quadlet like this feels...hacky. It works, but we've got a whole HEREDOC in the Justfile. :woozy_face: 
2. Should we automatically install the `ollama` cli via brew for the user? I'm leaning like we should, but they might not have brew installed.

> I've tested the Justfile in isolation, but not as a part of the main build. I don't have my fork setup for automatic builds yet. :grimacing:


castrojo (approved):

castrojo:
Yeah we could just include the file somewhere on disk and copy it over, it's only a few bytes. 

Let's punt on the magic ollama installation, maybe we can echo "make sure you set up brew first and install ollama" or something.

anoldguy:
> Let's punt on the magic ollama installation, maybe we can echo "make sure you set up brew first and install ollama" or something.

[More than this?](https://github.com/ublue-os/bluefin/pull/1072/files#diff-67001f02c5fe1c5f13defd6ee3c64fbd868f6023cac7434c22755a2bf3b4d404R102).

castrojo:
Ah yes, that'll work. 😄 

Ok, I'm fine with this and ready to +1! Would like a quick review from someone else on the quadlet tho.

p5 (approved):
LGTM!  Nice one!

p5 (approved):
Re-approving after resolved the merge conflict

p5 (approved):
Approving again after fixing Justfile syntax

PR #1077, opened 2024-03-30
fix: remove binaries from image

castrojo:
These should be brew installed, leaving kind on the image though.


castrojo:
Actually we should leave atuin in, people already have it set up and it's no big deal, better than dealing with migrating them to something.

castrojo (commented):
fix: leave atuin on the image

m2Giles (approved):

Issue #1036, opened 2024-03-20
Benchmark failed to run


mapringg:
### Describe the bug

- ujust benchmark

Instead I got this error:
```
aborting: temp-path '.' must be readable and writeable
error: Recipe `benchmark` failed on line 55 with exit code 1
```

### What did you expect to happen?

I expected the benchmark to run for 1 minute.

### Output of `rpm-ostree status`

```shell
State: idle
AutomaticUpdates: stage; rpm-ostreed-automatic.timer: no runs since boot
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bluefin-dx-nvidia:latest
                   Digest: sha256:d926ea236354a21020716b1ae4461a91533f37cce4741f22a6f71663bfc35ca9
                  Version: 39.20240319.0 (2024-03-19T16:56:24Z)

  ostree-image-signed:docker://ghcr.io/ublue-os/bluefin-dx-nvidia:latest
                   Digest: sha256:029a6a0e16d3becf5a2a509e829da95bc5d577e23a42c0404aca60a4bfb36b09
                  Version: 39.20240305.0 (2024-03-05T04:54:42Z)
```

castrojo:
Running the stress command manually works so it must be something else.

tepene:
Running `just benchmark` works but running `ujust benchmark` does not

![image](https://github.com/ublue-os/bluefin/assets/54766327/118da32f-aa0b-4aca-af7e-0e66b19fa37d)

evanphilip:
The problem seems to be that the `--justfile /usr/share/ublue-os/justfile` argument in the [definition](https://github.com/ublue-os/config/blob/7ace438c76dba68533189bb8a77630d8cadf753d/build/ublue-os-just/ujust#L3) of `ujust` changes the working directory to  `/usr/share/ublue-os`, which does not have the permissions necessary for `stress-ng`. This is the expected behavior of `just` and it is stated [here](https://just.systems/man/en/chapter_32.html?highlight=directory#disabling-changing-directory190) in Just Programmer's Manual
> `just` normally executes recipes with the current directory set to the directory that contains the `justfile`. This can be disabled using the `[no-cd]` attribute.

We have two options.
1. Disable the change of directory using the `[no-cd]` attribute. This will execute `ujust benchmark` in the current directory.
2. Change the directory to `/tmp` within the recipe of `ujust benchmark`, which would execute it in a safe directory with the necessary permissions. Personally, I prefer this option.

[...]

PR #1086, opened 2024-04-04
fix: change working dir to `/tmp` for `stress-ng` in `ujust benchmark`

evanphilip:
Change the working dir to `/tmp` for `stress-ng` in `ujust benchmark` to prevent read/write permission error as in https://github.com/ublue-os/bluefin/issues/1036

Note: setting `--temp-path` in `stress-ng` is not sufficient because `stress-ng` will continue to use the current directory for I/O tests.


castrojo (approved):

castrojo:
Thank you!

PR #1097, opened 2024-04-07
chore(ci): split workflows by fedora version

bsherman:
This will setup bluefin to build each Fedora version in distinct workflows like akmods, main, hwe.


m2Giles:
Can we also split between Aurora and Bluefin and put an exclude on each other's files?

This should prevent rebuilding an Aurora image when updating a file exclusively used by Bluefin and vice versa. While we will drop 38 in a few weeks, we are currently building over 50 images at a time.

I like where this is heading.

m2Giles:
We don't need to introduce Aurora 38. It will be gone in a few weeks.

Are we still able to exclude certain combos?

Ex: Framework 40 will be handled by another image.

bsherman:
> We don't need to introduce Aurora 38. It will be gone in a few weeks.

removed
 
> Are we still able to exclude certain combos?
> 
> Ex: Framework 40 will be handled by another image.

Yes... the excludes still work the same way in the reusable build

bsherman:
This PR will need to be force merged... and then the bluefin repo settings will need to be changed to select the new required successful jobs... instead of just "Check all builds successful" it should be all of:

- "Check all aurora 39 builds successful"
- "Check all bluefin 38 builds successful"
- "Check all bluefin 39 builds successful"

castrojo (dismissed):

.github/workflows/reusable-build.yml
```diff
@@ -279,7 +282,7 @@ jobs:
         run: exit 0
 
   build_iso:
```
p5, line 284: `build_iso` also contains the matrix, so I think we should fix that before this PR is merged.
Otherwise we will build ALL images 5 times over every time a PR is merged if I am understanding this correct?

bsherman: Ok, reviewed the yaml there...

While I agree with you it needs to be fixed, the build_iso job only runs on testing branch, and it's already a problem as it doesn't support aurora... 

Per Discord chat with @m2Giles , we are thinking merge this now, and he will tackle the Build ISO workflow improvements.

m2Giles: Great catch.

p5 (approved):

bsherman:
I updated the repo branch protection settings so there should be no hiccups on auto-approval for merge queue

PR #1099, opened 2024-04-08
chore(aurora): Remove unused dconf configuration

EyeCantCU:
Remove everything related to dconf sans configuration for Ptyxis


castrojo (approved):
Lots of skew today so things will be busted, if you get a green passing one just force merge it.

PR #1106, opened 2024-04-08
fix(ci): fix variable names for gts/latest tagging

m2Giles:


bsherman (approved):
Thanks for catching my bug here! this looks correct to me.

PR #1108, opened 2024-04-09
fix(40-dx): temporarily removed devpod from 40

m2Giles:
Fedora 40 dropped openssl1.1, Remove devpod temporarily.


castrojo (approved):

p5 (approved):
DevPod is only removed from 40, where it was broken anyway.  39 and GTS users are unaffected.

LGTM!

PR #1116, opened 2024-04-10
chore(ci): Update to latest version of Build-Container-Installer and include building Aurora ISOs

noelmiller:
# Summary

This PR includes a number of things in it. 
1. Using the latest pinned version of Build Container Installer
2. Using a folder upload rather than individual file uploads
3. Creates split workflows to build ISOs

# Blockers

- [x] Secrets needed for Aurora Buckets
- [x] Test ISOs that are being built with main
- [x] Pinned version of Build-Container Installer needs to be released
- [x] Pin to versioned release
- [x] Test ISOs built with pinned release version


m2Giles:
We have our flatpak lists duplicated

castrojo:
Yeah and shouldn't we use the `silverblue` and `kinoite` instead of kde_flatpaks, gnome_flatpaks, and aurora_flatpaks so we match how we organize the rest of the repo?

noelmiller:
> We have our flatpak lists duplicated

I think those got reintroduced on a merge from my branch. Removed them.

noelmiller:
> Yeah and shouldn't we use the `silverblue` and `kinoite` instead of kde_flatpaks, gnome_flatpaks, and aurora_flatpaks so we match how we organize the rest of the repo?

I guess I would want a second opinion on this @m2Giles. I personally think it's clearer to have them delineated between as bluefin_flatpaks and aurora_flatpaks personally.

noelmiller:
I'm going to say this is ready for review. The main blocker for building aurora ISOs shouldn't affect this going in. We can always fix it if the bucket name for aurora needs to be adjusted.

m2Giles:
I'm impartial to the names for the flatpak dirs.

One reason to stay with the brand names is that we invoke the workflow with this names. But we already use the kinoite/silverblue names inside as variables.

NiHaiden (commented):
Looks great to me, thanks Noel!

m2Giles (approved):
Let's get Aurora Shipping!

PR #1125, opened 2024-04-12
chore(ci): Fix passing secrets as outputs

noelmiller:
It occurred to me when running the workflow that echoing out the secrets to outputs could potentially expose the secrets as environment variables in the workflow. 

This is not preferred and to be safe, I would rather define individual steps in the workflow that passes the correct secrets based on the `inputs.brand_name`

We can go back to the previous method if we have proof that it doesn't but I would rather not take the risk at the moment.

### DIFF
--- a/.github/workflows/reusable-build-iso.yml
+++ b/.github/workflows/reusable-build-iso.yml
@@ -179,38 +179,34 @@ jobs:
           compression-level: 0
           overwrite: true
 
-      - name: Determine R2 Secrets
-        id: r2-secrets
-        shell: bash
-        run: |
-          if [[ ${{ inputs.brand_name }} = "bluefin" ]]; then
-            RCLONE_CONFIG_R2_ACCESS_KEY_ID=${{ secrets.R2_ACCESS_KEY_ID }}
-            RCLONE_CONFIG_R2_SECRET_ACCESS_KEY=${{ secrets.R2_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-build-iso.yml
```diff
@@ -179,38 +179,34 @@ jobs:
           compression-level: 0
           overwrite: true
 
-      - name: Determine R2 Secrets
-        id: r2-secrets
+      - name: Upload ISOs and Checksum to R2 to Bluefin Bucket
```
p5, line 182: You can have this as one step by using GitHub Environments.
You can have a secret for "R2_ACCESS_KEY_ID" for Bluefin, and a separate value for Aurora.

noelmiller: Hi @p5, that sounds like a good idea! So each environment has it's own secrets and you pick that in the workflow?

p5: Exactly.  You select which environment you wish to use at the job level, and you store separate secrets for each environment.
https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#using-an-environment

You can also have repo-level secrets which are accessible from all environments

noelmiller: Sounds good. I think we should open an issue to do that and move the secrets to the correct environment. I don't have time to do that at the moment, is this PR OK with folks just to get ISOs building?

p5 (approved):

PR #1127, opened 2024-04-12
feat: Add initramfs generation

KyleGospo:
TODO: Copy files from `system_files/nvidia/shared/*` into the image only for Nvidia builds


build_files/dx/build-dx.sh
m2Giles, line 1: I think we still need image-info.

build_files/shared/initramfs.sh
m2Giles, line 1: We don't have clients enabled so this will fail

system_files/shared/usr/lib/dracut/dracut.conf.d/10-compression.conf
m2Giles, line 1: We should compare against uncompressed for boot time.

Nvidia likely needs to be compressed for the additional files but non-nvidia initramfs will only be about 40 MB larger.

build_files/dx/build-dx.sh
KyleGospo: Mistake, fixed

noelmiller:
We will need the watermark to be image specific since Aurora builds out of the same repo. Not sure the best way to accomplish that.

m2Giles:
Two ideas:

1. Have the watermark under specific flavor files.

2. Rename each to branded, depending on name of image rename or symlink to generic watermark name.

noelmiller:
@KyleGospo @m2Giles have we tested this with ISOs yet? I don't think it should affect the installation, but I am curious about the first startup after installation.

m2Giles:
The only testing was with Bazzite testing.

Also need to make sure all of the Nvidia stuff is cleaned up. It's still included in -main images.

KyleGospo:
@m2Giles we'll want to keep the GSP Firmware enablement on all images that aren't Nvidia

p5 (approved):

castrojo (approved):

m2Giles (approved):

PR #1131, opened 2024-04-13
fix: remove duplicate builds on merge queue and push

p5:
We are building and pushing to GHCR when a PR is added to a merge queue (`merge_group`), and also after the PR has been successfully merged into main (`push`).

To me, this seems like we are running a build when we do not need to, and unnecessarily taking up GHA runners.

This PR removes the builds after a commit has been merged to `main` and `testing` since these are ran during merge queue.


p5:
I've cancelled builds because they are not required for this PR.

m2Giles (approved):

p5:
Going to bypass branch protection here.  No need for new builds.

PR #1136, opened 2024-04-14
feat: split -dx from base

m2Giles:


m2Giles:
Currently, the two jobs are not running sequentially. Concurrency cancels one of the two jobs.

m2Giles:
Current roadblock for this idea.

-dx builds are dependent on base builds being successful and available. In a PR, we don't publish that intermediate base meaning that we cannot check if a change in the base builds will negatively affect a -dx build.

PR #1146, opened 2024-04-15
fix: change switch-input-source keybinding

p5:
This `<Super>Space` keybinding should be used for the Search Light extension since it matches Alfred (MacOS).  
This PR is untested, and I will not be around to test it until tomorrow.

### DIFF
--- a/system_files/silverblue/usr/etc/dconf/db/local.d/01-ublue
+++ b/system_files/silverblue/usr/etc/dconf/db/local.d/01-ublue
@@ -35,6 +35,7 @@ switch-applications-backward = ['<Shift><Super>Tab']
 switch-windows = ['<Alt>Tab']
 switch-windows-backward = ['<Shift><Alt>Tab']
 switch-input-source = ['<Shift><Super>Space']
+switch-input-source-backwards = []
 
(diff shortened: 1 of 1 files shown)


castrojo (dismissed):

lauretano:
one thing to note, there's switch-input-source and switch-input-source-backwards, this has the effect of setting them to be the same shortcut. This doesn't seem to cause any problem since either shortcut brings up the switching OSD, and this still has the effect of freeing up super+space to be used by the search light extension. The "backwards" shortcut could be removed to clear up any potential confusion.

p5:
> one thing to note, there's switch-input-source and switch-input-source-backwards, this has the effect of setting them to be the same shortcut. This doesn't seem to cause any problem since either shortcut brings up the switching OSD, and this still has the effect of freeing up super+space to be used by the search light extension. The "backwards" shortcut could be removed to clear up any potential confusion.

@lauretano 
Good call!  Thanks!
I have made the change which I believe unsets that `-backwards` keybinding, but as with the other change in this PR, I am not in a position to test it.

m2Giles (approved):

PR #1148, opened 2024-04-16
fix: remove switch-input-source keybinding

p5:
This keybinding should be reserved for Search Light, so we need to unset the existing default usage of the keybind.

This is the second attempt at this, since the first attempt broke all dconf settings - restoring the system to a near stock GNOME shell.


m2Giles (commented):
Empty string is disabled?

lauretano (changes requested):
my error on the earlier pull request, the key name is switch-input-source-backward not backwards

castrojo:
This ready for review?

p5:
> This ready for review?

Give me 5 minutes to try to answer @m2Giles 's question, then will make it ready.  `dconf watch` isn't too helpful here, as it says it's just `[]`

p5:
Applied these dconf settings on a clean install of Bluefin, and this is the end result, which is expected.

image


Unfortunately Search Light is not working right now, but maybe a reboot will re-apply the shortcuts?

m2Giles (approved):

PR #1155, opened 2024-04-18
fix: source systemd-dx.sh from build-dx.sh script

p5:
I feel this file should be sourced (and therefore executed) from the build-dx.sh file.  This is why we have no Podman or Docker sockets enabled in DX.


bketelsen (approved):
thanks for catching this.

PR #1157, opened 2024-04-19
fix: restore v4l2loopback module

m2Giles:


bsherman (commented):
Why not installing v4l2loopbak on surface|asus? I think you can it was just not getting built for a short while when they were on 6.8 early.

castrojo (approved):

PR #1161, opened 2024-04-19
feat(ci): add image scanning workflow

p5:
Creates a reusable workflow to scan images.  This scan generates a SBOM and checks that SBOM for known vulnerabilities.

The only way I was able to merge the outputs of each matrix run was to use artifacts.  It's certainly not clean, but it works.

We needed to exclude some directories during the SBOM generation because Syft crashes when scanning too large images.

This PR only enables the scans for Bluefin until we finalise it.  Once we confirm everything is working, I will apply the same to Aurora and possibly Bazzite.


.github/workflows/reusable-build.yml
```diff
@@ -188,98 +194,123 @@ jobs:
...
+              exit 1
+            fi
+          done
+
+      - name: Create output
+        id: generate-outputs
+        env:
+          JOBS: ${{ toJson(needs) }}
+        run: |
+          images=$(echo $JOBS | jq -c '[to_entries[] | .value.outputs.image_full]')
```
p5, line 308: Current problem is the `toJson(needs)` does not contain information about all matrix jobs.  Only the one that finished most recently.

Will change each job to upload to a common artifact, and extract that in the final outputs step.  It's hacky, but should work.

p5: Of course that didn't work.  GitHub changed the way the upload-artifact action works in the recent v4 release.

Next idea is to create a job before the container builds which generates the matrix and sets the outputs. So set the output before fanning out to each individual build.

.github/workflows/reusable-image-scan.yml
```diff
@@ -0,0 +1,73 @@
...
+          curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
+          syft version
+
+      - name: Generate SBOM
+        env:
+          IMAGE: ${{ matrix.image }}
+        run: |
+          syft ${IMAGE} \
+            --output cyclonedx-json=sbom.json \
+            --config ./.github/syft.yml \
```
p5, line 49:
```suggestion
            --config ./.github/syft.yml
```

.github/workflows/reusable-image-scan.yml
p5, line 1: IMO we should centralise this configuration and reference it for all repos that want scanning.  There's nothing bluefin-specific in here.

.github/workflows/reusable-image-scan.yml
```diff
@@ -0,0 +1,73 @@
...
+          curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
+          syft version
+
+      - name: Generate SBOM
+        env:
+          IMAGE: ${{ matrix.image }}
+        run: |
+          syft ${IMAGE} \
+            --output cyclonedx-json=sbom.json \
+            --config ./.github/syft.yml
```
p5, line 49: We cannot (yet) use the action because it doesn't have support for passing in a config file.  
I have already created a PR for this, so hopefully we can switch soon!
https://github.com/anchore/scan-action/pull/299

.github/workflows/reusable-image-scan.yml
m2Giles: I think we can start here as a proof of concept.

We could then move it as it's own repo and have it be a callable action if we genericize it enough.

p5: Yeah, I plan to add things like uploading the SERIF files to GitHub, upload to R2 etc.  But that all comes later.  Right now, I just want to build the files and push them to job artifacts

m2Giles (approved):

PR #1164, opened 2024-04-21
fix: make sure quadlet is running before launching via shortcut

m2Giles:


system_files/silverblue/usr/etc/dconf/db/local.d/01-ublue
```diff
@@ -74,12 +74,12 @@ name='mission-center'
 
 [org/gnome/settings-daemon/plugins/media-keys/custom-keybindings/custom3]
 binding='<Control><Alt>Return'
-command="ptyxis --new-window"
+command='/usr/libexec/distrobox-quadlet-ptyxis.sh bluefin-cli'
```
castrojo, line 77: Doesn't this imply using the bluefin-cli box?

m2Giles: This does, we specify that the default is the Host here:
https://github.com/ublue-os/bluefin/blob/d45e53f4043afd128fd66f6afe681948d8b8558f/system_files/silverblue/usr/etc/dconf/db/local.d/01-ublue#L135

So previously this was just opening on the host if the user hadn't swapped over to bluefin-cli

castrojo (approved):

PR #1206, opened 2024-04-25
feat: add configure-grub just command

sekalengrengginang:
based on [bazzite configure grub just command](https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/share/ublue-os/just/80-bazzite.just#L44) with command from [hidden grub changeset](https://fedoraproject.org/wiki/Changes/HiddenGrubMenu)

### DIFF
--- a/system_files/shared/usr/libexec/configure-grub.sh
+++ b/system_files/shared/usr/libexec/configure-grub.sh
@@ -1,16 +1,26 @@
 #!/bin/bash
 source /usr/lib/ujust/ujust.sh
 GRUB_STATE="$(sudo grub2-editenv list | grep "^menu_auto_hide=")"
-if [ "$GRUB_STATE" == "menu_auto_hide=0" ] ; then
-    GRUB_STATE="${b}Not Hidden${n}"
+if [[ "$GRUB_STATE" == "menu_auto_hide=1" ]]; then
+    GRUB_STATE="${bold}Hidden${normal}"
+elif [[ "$GRUB_STATE" == "menu_auto_hide=2" ]]; then
+    GRUB_STATE="${bold}Always Hidden${normal}"
 else
-    GRUB_STATE="${b}Hidden${n}"
+    GRUB_STATE="${bold}Unhidden${normal}"
 fi
 echo "${bold}Grub menu configuration${normal}"
 echo "Grub menu is set to: $GRUB_STATE"
-OPTION=$(Choose "Hide Grub" "Unhide Grub" )
-if [[ "${OPTION,,}" =~ ^hide ]]; then
-    sudo grub2-editenv - set menu_auto_hide=2 boot_success=1
-    elif [[ "${OPTION,,}" =~ ^unhide ]]; then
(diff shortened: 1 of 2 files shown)


bketelsen:
i'll let others comment here too, but I think we're trying to reduce the amount of inline bash scripting in our just files and instead put an actual script on disk and call it from just.

sekalengrengginang:
> i'll let others comment here too, but I think we're trying to reduce the amount of inline bash scripting in our just files and instead put an actual script on disk and call it from just.

done, is /usr/libexec the correct location for the script?

system_files/shared/usr/libexec/configure-grub.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+source /usr/lib/ujust/ujust.sh
+GRUB_STATE="$(sudo grub2-editenv list | grep "^menu_auto_hide=")"
+if [ "$GRUB_STATE" == "menu_auto_hide=0" ] ; then
```
m2Giles, line 4: This isn't set on older installs. If the variable doesn't exist it is also not hidden.

system_files/shared/usr/libexec/configure-grub.sh
```diff
@@ -0,0 +1,16 @@
...
+    GRUB_STATE="${b}Not Hidden${n}"
+else
+    GRUB_STATE="${b}Hidden${n}"
+fi
+echo "${bold}Grub menu configuration${normal}"
+echo "Grub menu is set to: $GRUB_STATE"
+OPTION=$(Choose "Hide Grub" "Unhide Grub" )
+if [[ "${OPTION,,}" =~ ^hide ]]; then
+    sudo grub2-editenv - set menu_auto_hide=2 boot_success=1
```
m2Giles, line 13: menu_auto_hide=1 should hide the boot_menu so long as there was a successful shutdown.

Don't set boot_success=1. That variable should only be set when the computer successfully powers down.

m2Giles (changes requested):
Quick comments. menu_auto_hide=2 always hides. Additionally your condition check will say the grub menu is hidden even when it's not.

I think you should have 3 states.

Always Hide, Auto Hide, and Not Hidden. Auto Hide should be the default behaviour on new installs.

Thank you for adding this, and it will be a great place to put the rest of grub logic like removing duplicate entries and regenerating grub.cfg

system_files/shared/usr/libexec/configure-grub.sh
sekalengrengginang: it should exist since Fedora 29, did bluefin remove those env variable?

m2Giles: They weren't on my original 37 and 38 silverblue installs. Unsure why.

But it's besides the point if the variable doesn't exist it's not hidden which is what your code currently says.

m2Giles (approved):

sekalengrengginang:
thank you for adding check for grub state

Issue #1208, opened 2024-04-26
Ptyxis bluefin-cli profile breaks git commands requiring an SSH password


fbunt:
### Describe the bug

When using the bluefin-cli profile in Ptyxis, git commands that trigger SSH key password entry fail. I use a key that requires a password so `git clone` fails:

```sh
📦[fred@fedora ~]$ git clone git@github.com:fbunt/bluefin.git
Cloning into 'bluefin'...
ssh_askpass: exec(/usr/bin/ksshaskpass): No such file or directory
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

### What did you expect to happen?

I expected git commands that trigger password entry to succeed and provide a text or GUI password prompt.

### Output of `rpm-ostree status`

```shell
State: idle
AutomaticUpdates: stage; rpm-ostreed-automatic.timer: last run 8h ago
Deployments:
  ostree-image-signed:docker://ghcr.io/ublue-os/aurora-dx:latest
                   Digest: sha256:624d149bb1a62b9b6fd99c153b0f405b7494ad114df02039ae1ea5ab09dcacc9
                  Version: 40.20240425.0 (2024-04-25T16:51:42Z)
                     Diff: 1 upgraded
          LayeredPackages: micromamba sway swaybg swayidle swaylock

● ostree-image-signed:docker://ghcr.io/ublue-os/aurora-dx:latest
                   Digest: sha256:624d149bb1a62b9b6fd99c153b0f405b7494ad114df02039ae1ea5ab09dcacc9
                  Version: 40.20240425.0 (2024-04-25T16:51:42Z)
          LayeredPackages: micromamba sway swaybg swayidle swaylock

  ostree-image-signed:docker://ghcr.io/ublue-os/aurora-dx:latest
                   Digest: sha256:b64ef691153b101b3b4255091a0aecebd34f3ad647dec7de07b0036dd51a2202
                  Version: 39.20240419.0 (2024-04-19T19:43:17Z)
          LayeredPackages: micromamba sway swaybg swayidle swaylock
```

### Extra information or context

This is happening because the following variables are set:

```sh
📦[fred@fedora ~]$ env | grep ASKPASS
SSH_ASKPASS=/usr/bin/ksshaskpass
SSH_ASKPASS_REQUIRE=prefer
SUDO_ASKPASS=/usr/bin/ksshaskpass
```

[...]

PR #1210, opened 2024-04-26
fix: Do not set SSH_ASKPASS_REQUIRE in Aurora

m2Giles:
#1208

SSH_ASKPASS should only be prompted when the user does not have a terminal. Otherwise, if /usr/bin/ksshaskpass isn't available it won't fall back to the terminal for input. To trigger ksshaskpass from the cmdline, use `ssh-add < /dev/null`


p5 (approved):
Not entirely sure what this does, but it doesn't look bad

PR #1215, opened 2024-04-27
fix(bluefin): remove mutter patch on 40

m2Giles:


bsherman (approved):
:100: This is good. The patched mutter is nice to add scaling, but it's not worth waiting for the builds to be green again.

Plus, Bazzite already dropped this, at least temporarily.

PR #1216, opened 2024-04-27
feat: Replace ls with eza

dylanmtaylor:
image


m2Giles:
We're looking into implementing this via brew/fleek/sysext instead of layering so that we can have more control. We've previously had some of these utilities on image and quickly saw that there were some differences that were undesirable to setting an alias at that level.

PR #1224, opened 2024-04-27
feat: add Ollama WebUI Just script

p5:
Creates the just script for setting up an Ollama WebUI quadlet.
We needed to implement a custom network to bypass the recent Pasta 🍝 Podman networking limitations, where it discourages you putting stuff on the host network.


m2Giles:
On F40, podman is now podman 5 and defaults to Pasta for it's network backend. This is a breaking change from slirp4netns. 

The openwebui documentation calls out the necessary changes. I had to add the TCP port for the ollama quadlet and a host entry for it to work.

ericcurtin:
open-webui works fine for me, but I'm still on `Fedora 39` and `podman version 4.9.4` though

https://github.com/ericcurtin/podman-ollama

p5:
> open-webui works fine for me, but I'm still on `Fedora 39` and `podman version 4.9.4` though
> 
> https://github.com/ericcurtin/podman-ollama
> 

Yeah.  I need to add some extra args when using the default network, but when using a custom network, it works fine.  Will push the required args tonight

castrojo (approved):

ericcurtin:
Upgraded just to see if I could reproduce on Fedora 40 and podman 5.

It worked, but I think you are right, it's because I set up a pod with it's own custom network

PR #1226, opened 2024-04-28
feat: add kvmfr module (looking-glass) and vfio support

HikariKnight:
This adds support for the [kvmfr module](https://looking-glass.io/docs/rc/ivshmem_kvmfr/) used in [LookingGlass](https://looking-glass.io) and vfio support for VMs.
I have been testing this in bazzite since we got it added with Bazzite3.0 (Fedora40) and i have tried my best to make the module crash (and failed) in order to test it's stability.

This is used for GPU passthrough to windows VMs for those with supported systems that want a more seamless, no compromises solution for compatibility with some software (or for some games).
The LookingGlass team is looking to deprecate (and potentially remove) the support for the old shm method which atomic desktops relied upon before to have this work without a kernel module, hence the addition of this module.
image

This only enables the modules and does not fully configure everything for the user (no way we will do that), the user has to bind the card to the vfio-pci driver, configure the VM and compile `looking-glass-client` themselves (we in bazzite provides a guide for the latter that is linked to in the just recipe)

The LookingGlass team were ok with us including the `kvmfr` module as long as we at least took responsibility for our module package (as they had no interest in supporting "distro specific packaging") and filter out the issues that could be caused by our kernel module package (essentially, if user gets an image from the VM, the issue is upstream, otherwise it might be the module package, in which case it is my job to fix it. Which i am fine with as i made the akmods package the same way that `dkms` would. So they are 1:1)

~~# Needs clarification before merge
I am unsure if you want this in base or just dx so the PR right now have this added to base (since i figured some people might end up layering just virt-manager and qemu instead of using dx?), but i have no problems with moving it into dx only and i have the code ready for it.~~


m2Giles:
Only for -dx would be preferable since we have libvirt and qemu enabled there. I would not want to encourage a user to layer all of the qemu when we have the images. That said, containerized kvm could take advantage of this module.

Additionally, with the reorganization of akmods, we should move kvmfr from extras to common.

HikariKnight:
will switch to dx @m2Giles 
also i might be doing something wrong with the inclusion since i am getting an error about missing one of the dependencies for kvmfr akmod when building that we do not get in bazzite :thinking:  so i am investigating that too

m2Giles:
I would also prefer this to be a script and not inlined in just.

For the script check if we're on -dx otherwise recommend the user to rebase to -dx for virtualization features.

Additionally we can add the vfio drivers to the initramfs. I'm fairly certain they won't cause any issues being there. Then this should all be configurable with only kargs

HikariKnight:
> I would also prefer this to be a script and not inlined in just.
> 
> For the script check if we're on -dx otherwise recommend the user to rebase to -dx for virtualization features.
already adding a devmode check to the just, i can split the selinux and kvmfr configuration part into a script since that is where most of the "mess" is

> Additionally we can add the vfio drivers to the initramfs. I'm fairly certain they won't cause any issues being there. Then this should all be configurable with only kargs
The only reason we have kvmfr not configured in bazzite is due to handhelds. Plus upstream everywhere has vfio disabled by default.

m2Giles:
> The only reason we have kvmfr not configured in bazzite is due to handhelds. Plus upstream everywhere has vfio disabled by default.

Is there an issue with this? Arch and Proxmox have a few the modules in kernel and not as modules. Just wondering what the negatives would be.

HikariKnight:
> > The only reason we have kvmfr not configured in bazzite is due to handhelds. Plus upstream everywhere has vfio disabled by default.
> 
> Is there an issue with this? Arch and Proxmox have a few the modules in kernel and not as modules. Just wondering what the negatives would be.

I am guessing there is a chance that some oddball hardware might not like it. :shrug:
for me it does not matter, we can decide once i push the commit that splits the kvmfr part into a script

HikariKnight:
commit might fail with the akmods part, first time i have actually added an akmods so i just copied what kyle did in bazzite

system_files/dx/usr/bin/bluefin-dx-kvmfr-setup
m2Giles, line 1: Can you put this in libexec so it's not on $PATH

HikariKnight: will do

m2Giles (approved):

PR #1232, opened 2024-04-28
feat: added install-google-chrome && enable-luks-autounlock

rubixcube199:
google-chrome: https://github.com/bsherman/ublue-custom/blob/main/install-google-chrome.sh

enable-luks-autounlock: https://github.com/rubixcube199/OS-Setups-Configs/blob/main/laue.sh

### DIFF
--- a/just/bluefin-apps.just
+++ b/just/bluefin-apps.just
@@ -120,3 +120,127 @@ install-incus:
         echo "Run \"just devmode\" to turn on Developer mode."
         exit
     fi
+# Install Google Chrome # I have a desire to make a remark here.
+install-google-chrome:
+    #!/usr/bin/env sh
+
+    # Thanks to bri for the inspiration! My script is based on this example:
+    # https://github.com/briorg/bluefin/blob/c62c30a04d42fd959ea770722c6b51216b4ec45b/scripts/1password.sh
+
+    set -ouex pipefail
+
+    echo "Installing Google Chrome"
+
+    # On libostree systems, /opt is a symlink to /var/opt,
+    # which actually only exists on the live system. /var is
+    # a separate mutable, stateful FS that's overlaid onto
+    # the ostree rootfs. Therefore we need to install it into
+    # /usr/lib/google instead, and dynamically create a
+    # symbolic link /opt/google => /usr/lib/google upon
+    # boot.
+
+    # Prepare staging directory
+    mkdir -p /var/opt # -p just in case it exists
+
+    # Prepare alternatives directory
+    mkdir -p /var/lib/alternatives
+
+    # Setup repo
+    cat << EOF > /etc/yum.repos.d/google-chrome.repo
+   [google-chrome]
+   name=google-chrome
+   baseurl=https://dl.google.com/linux/chrome/rpm/stable/x86_64
+   enabled=1
+   gpgcheck=1
+   repo_gpgcheck=1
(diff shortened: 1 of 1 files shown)


just/bluefin-apps.just
```diff
@@ -120,3 +120,127 @@ install-incus:
         echo "Run \"just devmode\" to turn on Developer mode."
         exit
     fi
+# Install Google Chrome # I have a desire to make a remark here.
+install-google-chrome:
+    #!/usr/bin/env sh
```
m2Giles, line 125: It would be far preferable to have these as called scripts instead of inlining them inside the justfile.

This is advantageous since justfiles have strict formatting requirements which can break things.

just/bluefin-apps.just
```diff
@@ -120,3 +120,127 @@ install-incus:
...
+   # Clean up the yum repo (updates are baked into new images)
+   rm /etc/yum.repos.d/google-chrome.repo -f
+
+   # And then we do the hacky dance!
+   mv /var/opt/google /usr/lib/google # move this over here
+
+   #####
+   # Register path symlink
+   # We do this via tmpfiles.d so that it is created by the live system.
+   cat >/usr/lib/tmpfiles.d/google.conf <<EOF
```
m2Giles, line 177: this can't be written to in a booted system. The system admin editable location is /etc/tmpfiles.d/

just/bluefin-apps.just
```diff
@@ -120,3 +120,127 @@ install-incus:
...
+
+   ## Backup the crypttab
+   if [ -f /etc/crypttab.known-good ]; then
+     echo "Our backup already exists at /etc/crypttab.known-good\nExiting..."
+     [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
+   fi
+   cp -a /etc/crypttab /etc/crypttab.known-good
+
+   ## modify the crypttab
+   grep tpm2-device /etc/crypttab > /dev/null
```
m2Giles, line 209: this is not necessary for the root disk.

just/bluefin-apps.just
```diff
@@ -120,3 +120,127 @@ install-incus:
...
+
+   ## Update initramfs to respect tpm2 unlock
+   rpm-ostree initramfs |grep tpm2 > /dev/null
+   if [ 0 -eq $? ]; then
+     echo "TPM2 already present in rpm-ostree initramfs config."
+     rpm-ostree initramfs
+     echo
+     echo "Re-running initramfs to pickup changes above."
+   fi
+   rpm-ostree initramfs --enable --arg=--force-add --arg=tpm2-tss
```
m2Giles, line 236: we already have this dracutmodule inside of our initramfs.

m2Giles (changes requested):

p5 (commented):
The script you have copied from is meant to be called at build time (`docker/podman build`), not on a live system.  

A live system is far easier to install Chrome since it's almost identical to any Fedora distro

castrojo:
I'm a -1 on adding a layered browser, but let's at least echo a warning that it's an unsupported configuration with a confirm dialog or something (similar to homebrew)

noelmiller:
I am also -1 on layering a browser. The reason is browsers need to be updated on a regular basis and you will only get updates whenever the image is built. Flatpaks can be updated much more quickly and would not require the user to reboot.

PR #1243, opened 2024-05-01
chore(bluefin): match fedora logo names for system settings.

m2Giles:


dylanmtaylor:
The single color fedora logo looks great in system settings. Any chance we can do similar?

dylanmtaylor:
> This is to change that logo to bluefin with a dino

Right, I think that we should perhaps consider it being a single color in system settings like the original. It looks really clean that way.

dylanmtaylor:
Perhaps monochrome?
image
fedora_whitelogo

dylanmtaylor:
Or maybe just the text mark, with toned down colors.

image
fedora_whitelogo

PR #1251, opened 2024-05-03
feat: improve devcontainer experience

m2Giles:


m2Giles:
The Build ISO scripts are still WIP. Eventual goal is to add in docker qemu to launch and install those test ISOs.

Currently: just commands to build all flavors. Also the ability to run these containers with a quasi-booted like environment if using docker/rootful podman. There is still more to investigate with the booted like environments since it will consume your current environment. (You can vt-switch back, but it's bizarre).

Also, this now expects to use docker containers by default and is setup for docker-outside-of-docker for the quasi-booted like environment to work.

castrojo (approved):

PR #1256, opened 2024-05-05
fix: update search-light default appearance

p5:
I found the default search light a little hard to read in some scenarios, so this PR darkens the background, rounds the edges to look more gnome-like and adds a very thin border.

Unfortunately blur was moved to the testing package since it caused a few bugs, so we are unable to use that.  

Updates Search Light's default appearance from:
image

To:
image


m2Giles (approved):

castrojo (approved):

storopoli:
In `bluefin-dx-nvidia:gts` I am getting:

```bash
TypeError: background is null
```
image

PR #1264, opened 2024-05-07
feat: Add back xwayland video bridge to all images

m2Giles:
I'm unsure why we removed xwaylandvideobridge from Bluefin.


castrojo (approved):
I think we removed it because it was busted for a while. Should be good now!

PR #1265, opened 2024-05-07
feat: add bootc

m2Giles:
This is missing on Bluefin Latest.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4,6 +4,7 @@
 			"all": [
 				"bash-color-prompt",
 				"bcache-tools",
+				"bootc",
 				"evtest",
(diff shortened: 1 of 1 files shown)


p5:
We're only installing this in 39 for some reason.  @castrojo is that expected?  Or should it be on all versions?
https://github.com/ublue-os/main/blob/aec2dd03bbd6a940580870f8552e5c9821c52abc/packages.json#L197

m2Giles:
I think that 39 only was a holdover because of version differences between 38/39.

m2Giles:
Closing here and instead having this in main. https://github.com/ublue-os/main/pull/568

PR #1275, opened 2024-05-08
feat(framework): Use older mt7921e firmware

KyleGospo:


castrojo:
This is so awful, I love it!

castrojo (approved):

PR #1282, opened 2024-05-10
feat(framework): Stop enabling power-profile-switcher in ublue-user-setup

### LINKED ISSUE #1276: Framework power management wedges to Battery Saver when unplugged
### Describe the bug

I got a Framework 13 Ryzen yesterday and placed Bluefin DX 40 on it.

I have noticed that when the laptop is plugged in, I can move between power management settings (Performance/Balanced/Battery Saver). When it's unplugged, changing the setting doesn't do anything: it'll revert back to Battery Saver.

Power management when unplugged should respect what I am telling it and not switch straight back to Battery Saver. Let me have the juice!

_No response_

### Output of `groups`

_No response_

_No response_

cflewis:
power-profile-switcher is a good extension to install, but it's not a great default to start with; see the attached issue.

Remove enabling the switcher during user setup.

Fixes #1276

### DIFF
--- a/system_files/shared/usr/libexec/ublue-user-setup
+++ b/system_files/shared/usr/libexec/ublue-user-setup
@@ -44,8 +44,6 @@ fi
 
 if [[ ":Framework:" =~ ":$VEN_ID:" ]]; then
   if [[ ! -f "$UBLUE_CONFIG_DIR/framework-initialized" ]]; then
-    echo 'Enabling automatic power profile extension'
-    gnome-extensions enable power-profile-switcher@eliapasquali.github.io
     echo 'Setting Framework logo menu'
     dconf write /org/gnome/shell/extensions/Logo-menu/symbolic-icon true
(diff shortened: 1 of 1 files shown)


castrojo (approved):

cflewis:
My eyeballing of the checks is that they are transient 503 errors pulling the blobs. @castrojo : I you need to give it another kick :)

castrojo:
Enough of them passed, wanted to merge this to catch the cronned dailies, thanks!

castrojo:
Ok so we added some notifications to make it clearer that this is a managed configuration, but unfortunately it's non trivial for us to modify the menu to make a checkbox like "[ ] Automanage power" or something but this will at least give people breadcrumbs that it's managing your power. I'll add this to the docs so it's clearer.

It's not ideal but at least gets us some feature parity with KDE, maybe someday the GNOME Power Manager will be able to manage power. 😄

PR #1285, opened 2024-05-10
feat: add bootc to Bluefin

noelmiller:
# Upstream Issue
https://github.com/ublue-os/main/issues/572

# Purpose

We would like to test adding bootc to bluefin before adding it to main. Ideally, we should be able to test it here and then implement in main. 

# Things that need to change before merging

- [x] Putting booupctl command before initramfs step to see if it needs to be run before or after. This will help determine where this command needs to run (either in main or in the downstream)
- [x] Remove sl package
- [x] Test enough to get a release of build-container-installer so we aren't using the main branch


noelmiller:
@m2Giles feel free to work from this PR to get this merged into Bluefin.

m2Giles:
With the iso justfiles merged, we should be able to iterate this quickly. Will look at this later.

m2Giles:
Moving the bootupctl command before the initramfs generation worked.

m2Giles:
Note: using testing branch for the better brew install as well.

m2Giles:
Probably should of just rebased to before the brew stuff. But it's all removed.

This merge is currently waiting on a pinned release of the container build installer action.

gallium-stanza:
`bootc install` requires _gdisk_ which is not available in the image,
I tested installing on my machine with tpm2 encryption and it worked beautifully: `sudo podman run --rm --privileged --pid=host -v /var/lib/containers:/var/lib/containers -v /dev:/dev --security-opt label=type:unconfined_t ghcr.io/gallium-stanza/bluefin:testing /bin/bash -c 'rpm-ostree install gdisk --apply-live && bootc install to-disk /dev/nvme0n1 --wipe --filesystem=btrfs --block-setup=tpm2-luks'`

m2Giles:
I just realized I've only tested 40 images since we added the bootc.sh.

Need to make sure it doesn't negatively effect 39 images.

castrojo:
Ok so should I add gdisk in a separate PR?

m2Giles:
I think that is for the install to disk option of bootc.

Kinda like the coreOS installer command

m2Giles:
It looks like 39 doesn't like the bootupctl command at all.

image

geoffreysmith:
Do we have a slimmed down testing image for things like this (https://github.com/rust-vmm/rust-vmm-container/blob/main/Dockerfile)?) I am literally working on bootc testing because virt-vmm supposedly supports natively connecting to guest VMs, and supports booting directly from images. I guess just the fedora-bootc:41 ... maybe with nvidia because intercepts some things but I was hoping something lightweight. Their repo is like all over the place, I'm guessing this [](https://gitlab.com/fedora/bootc/base-images-experimental/-/blob/main/Containerfile?ref_type=heads)

build_files/base/bootc.sh
p5, line 1: This is in base and shared.  Here just doesn't contain the if condition.
Is the condition required - I.E. can this be ran on F39?

m2Giles: This cannot be run on F39. It will break the iso installer for F39.

It only needs to be in either base scripts. If you run the command more than once it will fail.

p5: Sorry, my point was `[build_files/base/bootc.sh](https://github.com/ublue-os/bluefin/pull/1285/files/b976975678d9e1906f574aa97baf23ca147bbde7#diff-100fefe85245ed5e3f8feeb529763e3cceb5538a729583b1faf3cc5e23574e6f)` runs it on all versions.  If this cannot be ran on F39, I think you need to reference the one in `build_files/shared/bootc.sh`?

p5: Oh, I don't know what I'm talking about.  This might be good

p5: Think this script might be better removed, since it's quite confusing

p5 (approved):
Thanks!  LGTM!
And sorry if it seemed like I was nit-picking!

m2Giles:
> Thanks! LGTM! And sorry if it seemed like I was nit-picking

Nope. This was weird where I had another branch where I was doing changes as well and some cleanup was missing. This is why we review!

m2Giles (approved):

PR #1289, opened 2024-05-11
feat: Restore power profile switcher on framework

KyleGospo:
Adds `ujust configure-auto-power-profile <option>` command
Switches to new version of power profile switcher modified with notifications when the profile is changed.


spacimek:
Is this only relevant for Bluefin or should be also ported to Aurora?

castrojo:
KDE has this out of the box, no need to do it there. This is mostly matching feature parity.

PR #1293, opened 2024-05-12
feat: homebrew on image

m2Giles:


m2Giles:
ToDo:

Better handling of permissions for lower to upper directory.

Right now the lower dir is owned by root. How do we make sure that brew can modify things without making duplicates of everything in the mutable upper directory. Or if duplication is fine getting that immediate chown fired.

But brew is on image.

m2Giles:
Root can use homebrew binaries if you have an interactive login shell.

For systemd services you can specify the binary with the full path. So something like rclone would be /home/linuxbrew/.linuxbrew/bin/rclone instead of /usr/bin/rclone.

m2Giles:
We'll need to adjust the brew just commands from config to not mess with this.

m2Giles:
Assumptions made:
1. The brew user is UID 1000. This makes sense for our systems:
    1. The Bluefin ISO will create the first user as a wheel user and UID 1000.
    1. The Aurora ISO needs the user to go into the advanced options to change the UID.
1. Single User system.
    1. Brew expects to be owned by a single user. Since most of our systems are single user our setup should be fine. However, for a multi-user system, UID 1000 needs to be the brew user.
1. Auto Updates
    1. Brew controls if it should auto-update based on an environment variable. Since brew and the portable ruby are on image, we do not want this to occur. This means that if a user goes awhile without rebooting, brew will not update.
    1. Need to inspect how auto upgrades work since normally you update prior to running upgrade.

system_files/kinoite/usr/share/ublue-os/motd/bluefin.md
```diff
@@ -6,7 +6,7 @@ Let's trace the stars.
 | ------- | ----------- |
 | `ujust --choose`  | Display available commands with search |
 | `ujust toggle-user-motd` | Toggle this banner on/off | 
-| `ujust brew` | Install Homebrew (Strongly Recommended) |
+| `ujust bluefin-cli` | Install Terminal Bling w/ Brew |
```
castrojo, line 9:
```suggestion
| `ujust bluefin-cli` | Enable terminal bling | 
```

system_files/silverblue/usr/share/ublue-os/motd/bluefin.md
```diff
@@ -5,7 +5,7 @@
 | ------- | ----------- |
 | `ujust --choose`  | Display available commands with search |
 | `ujust toggle-user-motd` | Toggle this banner on/off | 
-| `ujust brew` | Install Homebrew (Strongly Recommended) |
+| `ujust bluefin-cli` | Install Terminal Bling w/ Brew |
```
castrojo, line 8:
```suggestion
| `ujust bluefin-cli` | Enable terminal bling | 
```

just/bluefin-system.just
```diff
@@ -15,8 +15,13 @@ benchmark:
     echo 'Running a 1 minute benchmark ...'
     cd /tmp && stress-ng --matrix 0 -t 1m --times
 
-# Configure Bluefin-CLI Terminal Experience
+# Configure Bluefin-CLI Terminal Experience with Brew
 bluefin-cli:
+    brew bundle --file /usr/share/ublue-os/homebrew/bluefin-cli.Brewfile
+
```
castrojo, line 21:
```suggestion
bluefin-cli:
    brew bundle --file /usr/share/ublue-os/homebrew/bluefin-cli.Brewfile
    echo `Installation complete ... please close and reopen your terminal!`
```

castrojo (changes requested):

m2Giles:
Switching from the downloaded version of homebrew to the version installed by bluefin-cli, drops the duplicated data size down to 20 MB.

castrojo (approved):

m2Giles:
https://github.com/ublue-os/bluefin/pull/1301

This needs to be added.

PR #1309, opened 2024-05-16
fix(just): remove fleekbrew alias

castrojo:
`bluefin-cli` will do this by default, but not implemented yet so I'm just staging this here for now.


p5:
Are you intending on merging this immediately?  Or only when Bluefin CLI has it implemented?

castrojo:
Let's just clean this up now I think.

p5 (approved):

PR #1317, opened 2024-05-17
chore: stagger brew timers

m2Giles:
Have the two timers have 20 minutes of seperation.


system_files/shared/usr/lib/systemd/system/brew-update.timer
p5, line 1: Curious to know why the original configuration didn't work...

p5 (approved):

m2Giles (commented):
> Curious to know why the original configuration didn't work...  

The After line should work unless brew-update.service fails. Additionally, brew-upgrade doesn't rely on brew-update, it simply cannot run at the same time. Doing the time stagger I think is easier. Also means that brew-update can run more often while brew-upgrade can be slowed down.

PR #1347, opened 2024-05-25
feat: deprecate toolbox

castrojo:
40 only, 39/gts remains unchanged


p5 (approved):

NiHaiden (approved):

castrojo:
Ok so I asked around and enough red hatters use this that it's not worth making the change.  

We'll just continue to explicitly not recommend it.

PR #1351, opened 2024-05-27
feat(motd): add install date

castrojo:


noelmiller (approved):
LGTM, looks like the build failures are kernel skew.

PR #1360, opened 2024-06-01
chore(ci): remove SBOM and scanning workflows

p5:
These workflows sounded good in theory, but in reality they are too flakey and take far longer than I would have hoped to complete.  

I feel they are getting in the way more than they are helping.  Maybe later I will make them a workflow we trigger on an individual image manually (maybe put in ublue-os/main), but running every build doesn't seem useful.


castrojo:
I'd love to revist this in the future!

PR #1365, opened 2024-06-03
Fix broken path in brew setup script for fish

sqwxl:
The extension was missing from the filepath, which led to the completions script running on every shell startup because the `brew` symlink never exists. 

That completion script takes nearly a second to complete, and I found this bug when I was trying to understand why fish took so long to startup.


castrojo (approved):

castrojo:
Thank you!

PR #1368, opened 2024-06-05
feat: add skanpage and simple-scan

NiHaiden:
This PR adds the scanning apps skanpage for Aurora and simple-scan for Bluefin. Reason for layering directly is so the apps can access the drivers that are installed on the host system.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -44,6 +44,7 @@
 				"zsh"
 			],
 			"silverblue": [
+				"simple-scan",
 				"gnome-shell-extension-appindicator",
 				"gnome-shell-extension-blur-my-shell",
 				"gnome-shell-extension-caffeine",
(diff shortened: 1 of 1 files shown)


castrojo (approved):

salim-b:
@NiHaiden `simple-scan` is also available on Flathub as [`org.gnome.SimpleScan`](https://flathub.org/apps/org.gnome.SimpleScan). The Flatpak version ships the necessary drivers (`sane-backends` etc.), so I wonder whether anyone actually needs this app installed as RPM. Does Bluefin install additional scanner drivers that are missing from the Flatpak?

castrojo:
Yeah file this as an issue, we should swap it out for F42 along with some of the others.

PR #1372, opened 2024-06-07
feat: install flatpak-builder in dx

piperswe:
flatpak-builder can't easily be used in distrobox, and makes sense to have available in the developer version of a Flatpak-centric distro


castrojo (approved):
Thanks!

PR #1378, opened 2024-06-08
fix (ujust setup-cockpit): Fix PCP pmlogger permission and directories

lslowmotion:
This will allow users to have graph in Cockpit metrics and history.

pmlogger service won't run without the appropriate dirs and permission, too.
image


castrojo (approved):
This was annoying me but I didn't even know where to look, thanks so much!

PR #1380, opened 2024-06-09
feat(just): enable volume and add access info for ollama-web

lslowmotion:
Enabling volume will save user progress and account, so they wouldn't need to sign up over and over again after starting ollama-web systemd.

Also give user additional information regarding accessing Ollama Web UI.


castrojo (approved):
Thank you!

PR #1381, opened 2024-06-09
chore: make avif thumbnails work in dolphin

m2Giles:


KyleGospo:
Could this instead go in Main?

m2Giles:
> Could this instead go in Main? Seems like everybody would want this.

Yeah, that would be a great place for this. I'm guessing there is different packages for 39.

m2Giles:
https://github.com/ublue-os/main/pull/584

PR #1400, opened 2024-06-16
Added a .desktop shortcut to documentation

nicknamenamenick:
Links to https://universal-blue.discourse.group/docs?topic=2291


castrojo (approved):
Should we put this on the dock you think? cc @m2Giles @bketelsen

castrojo:
Maybe we can add it under "About my system" but above the next line so it's in the system section? 

image

m2Giles:
After about my system makes a ton of sense.

Also add a binding to maybe super+f1

PR #1402, opened 2024-06-16
feat: add openrazer-daemon

dreamyukii:
this pr will add openrazer-daemon to Bluefin and Aurora
i have 2 reason for adding this package 
1.Making easier for Razer peripheral users to configure their device since bluefin and aurora already including openrazer-kmod by default.
2. Configuration tools is available on flathub like [Razer Genie](https://flathub.org/apps/xyz.z3ntu.razergenie).


storopoli (changes requested):
That's too much bloat for the base image. I don't own any Razer things. I don't see why the user cannot overlay the package.

Maybe try to add it to the more gamer-oriented Bazzite images?

sekalengrengginang:
> That's too much bloat for the base image. I don't own any Razer things. I don't see why the user cannot overlay the package.
> 
> Maybe try to add it to the more gamer-oriented Bazzite images?

the problem with layering is that you need to remove the repo when upgrading to new version then re-add the repo since every Fedora version it will use different repo version.
Also bazzite already decided to make just recipe for layering openrazer-daemon.

storopoli:
> Also bazzite already decided to make just recipe for layering openrazer-daemon.

And why not proposing this same approach here?
A just recipe would not bloat the base image with things that a tiny minority of users need.

sekalengrengginang:
> > Also bazzite already decided to make just recipe for layering openrazer-daemon.
> 
> And why not proposing this same approach here? A just recipe would not bloat the base image with things that a tiny minority of users need.

i prefer this approach compared to layering, its the same as adding libratbagd daemon but for razer devices.
also it only add like 3-4 more rpm package since the rest of openrazer-daemon dependencies is already included

castrojo:
Not sure on this one still, let's see what others think.

m2Giles:
I think we should remove libratbagd. We probably should clean up old dependencies

Question, what does this daemon provide? Can this be run in a container?

castrojo:
I think it's vestigial and solaar replaces it? @KyleGospo would know.

m2Giles:
Yeah we also have vestigial stuff for the pano extension. We never ended up shipping it.

sekalengrengginang:
> I think we should remove libratbagd. We probably should clean up old dependencies
> 
> Question, what does this daemon provide? Can this be run in a container?

it's the same as libratbagd but for Razer devices, i tried to run this daemon with rootful distrobox but it failed to run

m2Giles:
We are removing the open Razer KMod because the mod pulls in Akmods as a whole (we do not need to include all srpm stuff on the image).

I'm very mixed on if we should have this kmod on the image especially if we do not include the userspace components.

dreamyukii:
I think removing hardware support and tell users to create custom image because their keyboard/mouse from Razer cannot be configured is not a good experience.

jmhbnz:
Hey team - I just stumbled across this in my travels of attempting to get my razer hardware working properly with bluefin. It looks like consensus hasn't been reached on merging this yet so I just wanted to ask is there a documented workaround somewhere I can follow on how to do this in a stock bluefin-nvidia fedora 41 install? Many thanks 🙏🏻

ledif:
Related: https://github.com/ublue-os/aurora/pull/44. Aurora merged this feature, but it requires some fixes to be functional.

dreamyukii:
> Related: [ublue-os/aurora#44](https://github.com/ublue-os/aurora/pull/44). Aurora merged this feature, but it requires some fixes to be functional.

Url for openrazer repo is updated and add systemd service for plugdev group

PR #1426, opened 2024-06-23
feat: Add ydotool to the dx image

dylanmtaylor:
With ydotool, users can script and automate inputs for mouse and keyboard in a way that works on Wayland.

https://github.com/ReimuNotMoe/ydotool

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -131,7 +131,8 @@
 				"ubuntu-family-fonts",
 				"udica",
 				"virt-manager",
-				"virt-viewer"
+				"virt-viewer",
+				"ydotool"
 			]
 		},
 		"exclude": {


castrojo:
This is pretty great, can you whip up quick instructions for the forum? Enabling one service is fine and probably not work doing a ujust for.

castrojo (approved):

dylanmtaylor:
> This is pretty great, can you whip up quick instructions for the forum? Enabling one service is fine and probably not work doing a ujust for.

I'm going to admit to something - I have it working and have tested it, but the way I did so is not 100% ideal. 

https://koji.fedoraproject.org/koji/buildinfo?buildID=2391183 says to run

```
- systemctl enable ydotool
- systemctl start ydotool
```

(This can be simplified to one command - `sudo systemctl enable ydotool --now`).

This was not sufficient for me as it ran as root and created a file at /tmp/.ydotool_socket owned by root. To make ydotool work without using sudo in front of it, I did this:

```
export YDOTOOL_SOCKET=/tmp/.ydotool_socket # You can put this in your profile files
sudo chmod 666 /tmp/.ydotool_socket
```

I know the latter command is not great for security. 

Someone at https://discussion.fedoraproject.org/t/ydotool-not-working-on-fedora-40/114333/4 had a better suggestion:

```
sudo dnf install ydotool
sudo tee /etc/profile.d/ydotool.sh << "EOF" > /dev/null
export YDOTOOL_SOCKET="/tmp/.ydotool_socket"
EOF
sudo mkdir -p /etc/systemd/system/ydotool.service.d
sudo tee /etc/systemd/system/ydotool.service.d/override.conf << EOF > /dev/null
[Service]
Group=input
ExecStart=
ExecStart=ydotoold -P 660
EOF
sudo systemctl daemon-reload
sudo systemctl enable ydotool.service
sudo systemctl restart ydotool.service
sudo usermod -a -G input ${USER}
```

If I want to recommend a solution to people other than myself, I'd want it to be more secure. The above is probably WAY better, and can be a ujust action.

dylanmtaylor:
To test this, here's a valid example:

```
ydotool type -D 1000 -d 50 'Universal blue is awesome!'
```

m2Giles:
In this case I think a ujust command makes more sense.

I'm unsure if the command accepts UID and GID for that socket creation.

A ujust command would make it possible to setup the overrides appropriately.

dylanmtaylor:
> In this case I think a ujust command makes more sense.
> 
> I'm unsure if the command accepts UID and GID for that socket creation.
> 
> A ujust command would make it possible to setup the overrides appropriately.

I agree with this. I'm not familiar with ujust, but will give it a shot. Can we merge this? I'll start work on the ujust in a separate PR.

dylanmtaylor:
Looks like asus kernel builds were pulled for 39. Maybe we need to add an override to not build those

https://download.copr.fedorainfracloud.org/results/lukenukem/asus-kernel/fedora-39-x86_64/repodata/repomd.xml

PR #1428, opened 2024-06-23
feat: Add ydotool setup script to ujust

dylanmtaylor:
Prerequisite: #1426 

I'm not confident this one is right, so marking as draft.


dylanmtaylor:
If you want me to do a forum post instead, that's fine. We can close this. I just took feedback from M2 and thought it'd be nice to make this easier to consume.

m2Giles:
Yeah, let's just have a forum post for this. That service unit has issues.

PR #1433, opened 2024-06-23
chore(ci): clean up workflow names

castrojo:

### DIFF
--- a/.github/workflows/build-coreos-bluefin.yml
+++ b/.github/workflows/build-coreos-bluefin.yml
@@ -1,4 +1,4 @@
-name: bluefin CoreOS Kernel
+name: CoreOS Kernel
 on:
   merge_group:
   pull_request:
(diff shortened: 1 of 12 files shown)


p5 (dismissed):

.github/workflows/build-coreos-bluefin.yml
m2Giles, line 1: Unless we aren't doing Aurora as well, we should probably have Bluefin in the title.

castrojo: Wouldn't it just be one workflow for both?

m2Giles: Unfortunately no. We have brand_name as an input which is how bluefin or Aurora is decided to be built.

Even as I think about GTS vs Latest, I may have to completely restructure the matrix for the reusable build to handle across versions.

m2Giles: I could possibly look into some sort of matrix option on our current workflow to add coreOS as a build option

castrojo: Ok I'll just wait post-coreos kernel and then reapply what makes sense.

m2Giles (approved):

PR #1434, opened 2024-06-24
feat: Set ELECTRON_OZONE_PLATFORM_HINT=auto as the default

polkaulfield:
Fixes rendering issues and crashing of electron apps when using the NVIDIA drivers. 

This makes the electron applications (flatpak or not) to detect the type of session and act accordingly.

https://www.electronjs.org/docs/latest/api/environment-variables#electron_ozone_platform_hint-linux


castrojo (approved):
Oooh, thanks!

polkaulfield:
From my testing everything gets fixed, the only downside is that with some apps, you get the wayland icon instead of the app one when alt-tabbing. Other than that, tested element, boxy-svg, signal, discord, spotify and most importantly VSCode and all the UI rendering bugs and stuttering go away :) 

Maybe if you prefer to use Xwayland it could be merged just for NVIDIA, but i think the Wayland icon is a small downside to the performance improvements.

polkaulfield:
I just checked it and it seems that /usr/etc/environment doesnt get sourced if you have changes already in /etc. Is that the expected behavior? I started using silverblue based Linux a few weeks ago.

p5:
Yes, unfortunately if somebody has edited their local file, the updates to that same file will not be applied.

We could look to change the configuration to /etc/environment.d/something.conf, which should help.  Not too sure if this is a valid drop-in directory yet

polkaulfield:
Ok! I'll work on it. Seems a way better approach. What about /etc/environment.d/electron-flags.conf?

PR #1438, opened 2024-06-24
Revert "feat: Set ELECTRON_OZONE_PLATFORM_HINT=auto as the default"

castrojo:
Reverts ublue-os/bluefin#1434

Kyle reporting Discord crashes, reverting for now.


polkaulfield:
Damn. Discord worked fine here. Should we add it just for the nvidia flavors?

polkaulfield:
I have been testing most of the commonly used apps and there's a crash with Slack too. With ELECTRON_OZONE_PLATFORM_HINT=x11 set up on flatseal both run fine.

KyleGospo:
> I have been testing most of the commonly used apps and there's a crash with Slack too. With ELECTRON_OZONE_PLATFORM_HINT=x11 set up on flatseal both run fine.

Requiring users to have to manually go set environment arguments to make their applications stop breaking as a terrible experience, we can't ship this.

polkaulfield:
Damn... You are absolutely right.

Im using NVIDIA and everything was unusable almost without this flag. Thought it wouldn't create problems using `auto` instead of forcing wayland.

I tested a bunch of electron apps but there are way too many to check i guess, and yeah Discord is absolutely broken, i tested with Webcord and probably had a brain fart thinking i did with the official Discord app.

Maybe we could add this workaround in the documentation as a manual fix using flatseal for individual apps instead of setting it globally.

I apology for the breakage it has caused :(

KyleGospo:
You're good, you can't catch everything and it was for a good cause. 

We cut it early so no harm.

polkaulfield:
I was thinking of adding the flag just on vscode in the nvidia dx builds. That one doesn't break from my testing, it's unusable without it. What do you think? I can open a PR with that change.

polkaulfield:
Yeah, I was thinking that maybe an ujust script that allows you to do it on a per app basis might be the way (that adds the flag and wayland permission to the chosen problematic apps). I found out that most of the crashes on flatpak apps can be prevented toggling the wayland permission on when having the flag enabled.

polkaulfield:
I'll write one and send a PR to the testing branch when i have time if you think it's a good idea.

castrojo:
Should we wait for the next nvidia driver drop? I'm cognizant of adding a bunch of skaffolding and then having to throw it away later.

polkaulfield:
Well, I wrote a proof of concept script that adds the tweak for the flatpaks you choose.
https://gist.github.com/polkaulfield/f1c83ca70735edbeed2d52c78e8c9248
Maybe it's a good approach to have an ujust script with that

PR #1456, opened 2024-06-27
feat: GTS use coreos matched kernel

m2Giles:
GTS will use the matched kernel of whatever is the current coreos version. Additionally, now uses the convenience tags instead of version numbers for determining which version of fedora to build. Stages having beta builds for F41. Deprecate aurora gts


m2Giles:
This is now green. Last bits are updating the just scripts and updating the README.md

m2Giles:
While Nvidia is being installed using the same method as bazzite/hwe a user is reporting an issue with driver mismatch.

m2Giles:
Nvidia issue was due to an update in between akmods build and the image build.

p5 (approved):
LGTM!

.github/workflows/reusable-build.yml
ethanjli, line 1: Question: in line 243 of this file (i.e. within the "$IS_COREOS" == "true" branch), there is no `BUILD_TAGS+=("stable")` line. Was this omission intentional?

In my fork of the bluefin repo (which is now just a layer of changes over Aurora) I found I had to [add that line](https://github.com/ethanjli/lutho-os/commit/b53d4b3d9fe109af71ea236f2735544474244e6c) in my copy of the reusable-build-iso workflow - without it, the "Build Image" step [errored out](https://github.com/ethanjli/lutho-os/actions/runs/9748427516/job/26903392849) failed because the `BUILD_TAGS` array was empty.

PR #1461, opened 2024-06-29
chore(ci): Switch GTS/Stable to only Scheduled builds. Build ISOs weekly for GTS with schedule

m2Giles:

### DIFF
--- a/.github/workflows/build-39-bluefin.yml
+++ b/.github/workflows/build-39-bluefin.yml
@@ -4,6 +4,9 @@ on:
     branches:
       - main
       - testing
+    paths-ignore:
+      - '**.md'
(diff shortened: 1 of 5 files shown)


.github/workflows/build-39-bluefin.yml
```diff
@@ -1,21 +1,11 @@
 name: Bluefin 39
 on:
-  merge_group:
   pull_request:
     branches:
       - main
       - testing
-    paths-ignore:
-      - '**.md'
-      - 'system_files/kinoite/**'
```
p5, line 10: I think you'll still want these paths-ignore so these do not build in PRs when you don't need them to?

p5 (approved):

PR #1477, opened 2024-07-02
feat: ujust command to toggle Tailscale

Raindrac:
This PR adds a ujust command to toggle the tailscaled service. Because the tailscaled service is active in the background even if the user hasn't set up Tailscale, it increases power usage and battery consumption regardless of whether Tailscale is in use. Users who don't want the service running in the background can now disable it, and this lays the groundwork for making Tailscale opt-in rather than opt-out.

Tested locally on Bluefin. This is my first ujust contribution and the first time I've wrote a bash script, so corrections are welcome.


Raindrac:
Converted to draft. Because only toggling the service might cause confusion, I'm going to make this command toggle the extension and hook it up to gum as well.

Raindrac:
I've tested the updated script on a local build. It now enables or disables the Tailscale QS extension for Gnome, and when Tailscale is enabled, it notifies the user that they may have to set it up and links them to the documentation. Ready for review!

storopoli:
That's great! Looking forward to having this merged.

Raindrac:
Alrighty, I'm looking into why the just build check is failing now 🤔

Raindrac:
That should work, I think! It passes the just check locally now. It was just a line ending 😞

just/bluefin-system.just
```diff
@@ -280,3 +280,49 @@ update-ng:
     sudo bootc upgrade
     flatpak update -y
     brew upgrade
+
+# Toggle tailscale
+toggle-tailscale:
+    #!/bin/bash
```
m2Giles, line 286: please use:

set -euo pipefail to failout in undefined situations.

just/bluefin-system.just
```diff
@@ -280,3 +280,49 @@ update-ng:
...
+
+    TAILSCALED_STATUS="$(systemctl is-enabled tailscaled)"
+
+    if [ "$TAILSCALED_STATUS" == "enabled" ]; then
+        TAILSCALED="Installed"
+    elif [ "$TAILSCALED_STATUS" == "disabled" ]; then
+        TAILSCALED="Installed"
+    else 
+        echo "${b}${red}Unable to enable or disable Tailscale.${n}"
+        echo "tailscaled service status: $TAILSCALED_STATUS"
```
m2Giles, line 297: This should be saying cannot find tailscaled.

Raindrac: Per the documentation of [is-enabled](https://www.man7.org/linux/man-pages/man1/systemctl.1.html), there are other potential output statuses that could be returned which I haven't tested for, some of which would be returned in scenarios where the unit file is present. So my concern is that stating "Cannot find tailscaled" might not necessarily be accurate, because sometimes the command could fail despite recognizing the presence of of tailscaled. My approach was to instead echo the service status returned by systemctl, informing the user of the status and allowing them to address the problem through methods such as unmasking the unit file.

Do you think I should rewrite the check, continue with changing the message to "Cannot find the tailscaled service.", or leave it as-is?

m2Giles: Leave as is then.

m2Giles (approved):

PR #1493, opened 2024-07-09
feat: Ship fsync kernel on latest, drop HWE builds

EyeCantCU:
Ships the fsync kernel on `:latest`. Due to the fsync kernel's broad hardware support, we are able to drop HWE builds as support for Asus and Surface devices is already implemented in the kernel


EyeCantCU:
Drafted until we've implemented a path forward for HWE users

m2Giles:
fsync cache package is going to change names to fsync-kernel. See https://github.com/ublue-os/kernel-cache/pull/5

Additionally, that pull request also caches the coreos kernel as well now. So we can do the same install method for our images.

m2Giles:
Fsync is now on latest for main/Nvidia.

Next step is removing hwe and having people rebased over.

Surface and Asus both have some extra packages to need to investigate what's going on with those.

Surface:
libwacom
libwacom-data
iptsd
Watchdog

Asus:
asusctl

m2Giles:
Fsync has been deprecated.

PR #1522, opened 2024-07-21
chore(ci): explicitly use v1.2.1 for ISO action

castrojo:


m2Giles (approved):
The hash matches the release, but we can switch to v1.2.1 for the time being.

PR #1532, opened 2024-07-23
fead: Add dialog boxes

lethedata:
Add zenity and kdialog dialog boxes for silverblue and kinoite respectively to support distrobox rootful container gui applications.


lethedata:
The specific reason for this is that `distrobox_sudo_askpass` is a file generated by `distrobox-export` which is for using commands other than sudeo to run a rootful container.[^1][^2]

[^1]: [distrobox-export#L54](https://github.com/89luca89/distrobox/blob/4c38d6b3f9bc833e58a435471158935824727805/distrobox-export#L54)
[^2]: [Using a command other than sudo to run a rootful container](https://github.com/89luca89/distrobox/blob/main/docs/useful_tips.md#using-a-command-other-than-sudo-to-run-a-rootful-container)

m2Giles:
Gnome-ssh-askpass and ksshaskpass are both present for a graphical prompt.

You can also use polkit as well with either systemd-run or pkexec.

I'm unsure how these would be necessary for the specified use case of rootful distroboxes.

That said, both of these can be used to make simple graphical prompts for scripts.

lethedata:
> Gnome-ssh-askpass and ksshaskpass are both present for a graphical prompt.
> 
> You can also use polkit as well with either systemd-run or pkexec.
>
> I'm unsure how these would be necessary for the specified use case of rootful distroboxes.

The `distrobox_sudo_askpass` script is created by `distrobox-export` by default and used to ask for a password on exported apps. It seems to be distrobox's default way to ask for a sudo password outside of a terminal. Using one of those mentioned above is an option it just requires some manual intervention.

> That said, both of these can be used to make simple graphical prompts for scripts.

I admit that this is probably better reasoning than distrobox specific. I mention distrobox because of above and it being bundled by default.

PR #1534, opened 2024-07-24
feat: assemble custom os-release file

castrojo:
To match bazzite


m2Giles:
We already have an image-info.sh. We should match behaviours between bazzite and us. Right now there are two files doing a similar thing.

Issue #1355, opened 2024-05-29
New dinosaur artwork for fastfetch


castrojo:
Three new dinos from Jacob. We use chafa to generate them for the terminals: https://hpjansson.org/chafa/

Let's add these to the rotation!

![Jorge_CustomChonks_Roaring](https://github.com/ublue-os/bluefin/assets/1264109/49c32d54-7f86-4780-98cd-6dd03be63e57)
![Jorge_CustomChonks_Leaping](https://github.com/ublue-os/bluefin/assets/1264109/1a41a5da-c203-4168-9e04-085f2565c5bc)
![Jorge_CustomChonks_Intrigued](https://github.com/ublue-os/bluefin/assets/1264109/d1afcc92-de16-4711-9f02-66aaaf65fd8e)

dertoni:
Hi,

so this is my first pull request, so bear with me :-)

Things done:

    Saved the three new dinosaur pngs under 'system_files/silverblue/usr/share/ublue-os/bluefin-logos/'
    I took the liberty of naming them ducky, luke and runner
    Generated the files under 'system_files/silverblue/usr/share/ublue-os/bluefin-logos/symbols/' with 'chafa > '
    Generated the files under 'system_files/silverblue/usr/share/ublue-os/bluefin-logos/sixels/' with 'chafa -f sixels >

Is that the right way? Is there anything else todo?

PR #1535, opened 2024-07-25
Added new dinosaur logos for fastfetch

dertoni:
Hi, 

so this is my first pull request, so bear with me :-)

This should address #1355 

Things done:
- Saved the three new dinosaur pngs under 'system_files/silverblue/usr/share/ublue-os/bluefin-logos/'
- I took the liberty of naming them ducky, luke and runner
- Generated the files under 'system_files/silverblue/usr/share/ublue-os/bluefin-logos/symbols/' with 'chafa <png> > <path>'
- Generated the files under 'system_files/silverblue/usr/share/ublue-os/bluefin-logos/sixels/' with 'chafa -f sixels <png> > <path>

Is that the right way? Is there anything else todo?


castrojo:
I'll try to get to this when I can but passing along my thanks in the meantime!

castrojo:
image

Ok I landed these in :latest, they look good but there's some extra whitespace which pushes the fastfetch stats to the right. Also wondering if we should start a discussion on the default font size and the fastfetches since ootb they don't really match. We're heading in the right direction!

PR #1537, opened 2024-07-27
chore: clean up vestigial flatpaks

castrojo:


m2Giles:
While vestigial I think one of our setup scripts reads them.

m2Giles (approved):

Issue #1547, opened 2024-07-29
Add Tailscale tray app for Aurora


RealVishy:
### Describe the package

I'd like to request one of the below packages for Aurora/Bazzite so it matches the tailscale QS shipped in gnome. I'll try them out and see which one i end up liking more.

https://github.com/DeedleFake/trayscale 

https://github.com/f-koehler/KTailctl

### Image

All Images

RealVishy:
KTailctl integrates better with KDE and it seems to have about the same features as trayscale so I think it's the better pick.

I'd like to submit a PR for it but i'm not sure how to add a flatpak to the images.

castrojo:
https://github.com/ublue-os/bluefin/blob/main/aurora_flatpaks/flatpaks

Great idea! After you PR let's give it a few days for people to weigh in, both of them appear to be maintained recently so that's always a positive sign. Thanks!

RealVishy:
one "caveat" is that for full functionality of the app, it requires the user to have run `sudo tailscale up --operator=$USER`.

I'm currently following the [existing issue](https://github.com/ublue-os/bluefin/issues/862) for it and i'll see if i can also merge the fix used in bazzite.

RealVishy:
Turns out someone already put in the fix from bazzite, closing the issue now.

PR #1548, opened 2024-07-29
add KTailsctl package

RealVishy:
Adds package required in this issue.

https://github.com/ublue-os/bluefin/issues/1547


castrojo:
Don't worry about the semantic check failing it's not required.

RealVishy:
i'm guessing this is blocked until the kernel cache thing in stable is fixed?

castrojo:
Yep, we only fixed that today but ran into some version skew from fedora. As soon as it's all green I'll merge this, thanks for your contribution!

Issue #1491, opened 2024-07-06
Add fish container indicator from bazzite


castrojo:
Filing as a reminder:

https://github.com/ublue-os/bazzite/pull/1337

RealVishy:
I'd like to help with this when I've got time 😄

RealVishy:
I'll test it out now that it's merged but I think starship might conflict with this fish function

RealVishy:
I believe the function assumes that you're using the same home directory for your distrobox for it to be able to load in the fish function. Probably needs a rename as well.

PR #1569, opened 2024-08-06
feat: container indicator for fish shell

RealVishy:
Copies over the fish function from bazzite.

Afaik it's the right folder to be put in but i'll fix it if i'm wrong.

https://github.com/ublue-os/bluefin/issues/1491


castrojo:
I don't use fish so I can't test this, anyone else out there using fish?

RealVishy:
apparently starship makes it's own function with the same name.

```❯ functions fish_prompt
# Defined in /tmp/.psub.Wj8KHNUFnO @ line 1
function fish_prompt
    switch "$fish_key_bindings"
        case fish_hybrid_key_bindings fish_vi_key_bindings
            set STARSHIP_KEYMAP "$fish_bind_mode"
        case '*'
            set STARSHIP_KEYMAP insert
    end
    set STARSHIP_CMD_PIPESTATUS $pipestatus
    set STARSHIP_CMD_STATUS $status
    # Account for changes in variable name between v2.7 and v3.0
    set STARSHIP_DURATION "$CMD_DURATION$cmd_duration"
    set STARSHIP_JOBS (count (jobs -p))
    if test "$TRANSIENT" = "1"
        # Clear from cursor to end of screen as `commandline -f repaint` does not do this
        # See https://github.com/fish-shell/fish-shell/issues/8418
        printf \e\[0J
        if type -q starship_transient_prompt_func
            starship_transient_prompt_func
        else
            printf "\e[1;32m❯\e[0m "
        end
    else
        /usr/bin/starship prompt --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="$STARSHIP_CMD_PIPESTATUS" --keymap=$STARSHIP_KEYMAP --cmd-duration=$STARSHIP_DURATION --jobs=$STARSHIP_JOBS
    end
end
```

castrojo:
It was poor form on my part to merge without giving you a heads up, so whatever changes you need to make it good I'll be standing by, appreciate your help!

RealVishy:
AFAIK just renaming the function should fix it but I'll read up on the fish documentation. No worries about it being merged early, I'll try to get it fixed soon!

RealVishy:
One of the reasons it wasn't working for me was that it needs your distrobox to be using the same home dir. I was using a custom home dir, which is why it wasn't loading in the fish function.

PR #1580, opened 2024-08-08
fix: Move files in /usr/etc to /etc

KyleGospo:
Adding files to /usr/etc is considered undefined behavior


system_files/shared/usr/lib/systemd/system/ublue-etc-merge.service
m2Giles, line 1: Probably should remove this file. This is for the whole takeover your computer with a docker container

castrojo:
Ok I've removed all mentions of the etc-merge service. Please doublecheck my work.

castrojo:
This good to go or are we still sorting out that containers.json thing? (Does that even affect Bluefin since we're not using the rechunker?)

p5:
> This good to go or are we still sorting out that containers.json thing? (Does that even affect Bluefin since we're not using the rechunker?)

@castrojo The containers.json thing is only affecting main and ucore, since this is where we add the container policy to the image.  The issue there is we want to move the policy from /usr/etc to /etc, but Fedora's containers package already has taken "ownership" of that file, and doesn't allow us to overwrite it from our own RPM. 
This has been reverted for now, but I don't think we have any issues here.

PR #1588, opened 2024-08-10
feat: ollama ujust script, some additions for removal verbs for the ollama and open-webui containers.

muddyfeather:
this is a first step on adding some actions to the ujust for ollama before potentially standardising this in bazzite too with another PR to bring "bazzite-tools.just" over. I've been using this, but i've seen others with the same opinion as me that i'd like to be able to mostly atomically clean up after it, as bazzite machines are often everything devices (so people are picky about things using their resources) I also can imagine a situation where in bazzite, people will want to tinker with this, but then remove and clear up afterwards because of the nature of the machine, and the users themselves are likely not as tech savvy as say aurora or bluefin people.
The script is a frankenstein now, but works in my test environment, backwards and forwards as necessary. I'm not sure what the established design conventions are for remove- verbs and how people expect it to work, but it's got an included "ollama help" to move things along. 

I'm sure this will help dozens of people ;)

### DIFF
--- a/just/bluefin-tools.just
+++ b/just/bluefin-tools.just
@@ -26,141 +26,182 @@ tensorflow:
     podman run -it -p 8888:8888 docker.io/tensorflow/tensorflow:latest-jupyter  # Start Jupyter server
 
 # Setup a local Ollama instance in a container. Detect hardware, offer a choice if needed.
-ollama:
+install-ollama ACTION="":
     #!/usr/bin/env bash
-    echo 'Detecting Hardware...'
-    echo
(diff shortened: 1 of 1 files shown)


muddyfeather:
FYI, happy to redo, if we prefer install-ollama, install-open-webui, remove-ollama recipes as a convention!

castrojo:
I'm hoping we can get some testing on this, I'm short on time lately. :(

castrojo:
The weekly builds just passed so let's just get this into latest. Thanks for helping with this!

muddyfeather:
No problems!

Issue #1596, opened 2024-08-12
Breeze-GTK is not included by default on Aurora


ThoriumTextile:
### Describe the bug

On any fresh install of Aurora, GTK apps will default to adwaita dark.

### What did you expect to happen?

GTK apps should use the Breeze theme, but this package is missing by default.

### Output of `rpm-ostree status`

```shell
State: idle
AutomaticUpdates: stage; rpm-ostreed-automatic.timer: last run 36min ago
Deployments:
  ostree-image-signed:docker://ghcr.io/ublue-os/aurora:stable
                   Digest: sha256:109a1480aefa1faef8f37b5de7ae143c2b2c961cd372c2415ec6fe7777944efe
                  Version: 40.20240807.0 (2024-08-07T13:32:18Z)
                     Diff: 1 upgraded
          LayeredPackages: breeze-gtk corectrl distrobox docker java-17-openjdk kontact micro wine-core

● ostree-image-signed:docker://ghcr.io/ublue-os/aurora:stable
                   Digest: sha256:109a1480aefa1faef8f37b5de7ae143c2b2c961cd372c2415ec6fe7777944efe
                  Version: 40.20240807.0 (2024-08-07T13:32:18Z)
          LayeredPackages: breeze-gtk corectrl distrobox docker java-17-openjdk kontact micro wine-core

  ostree-image-signed:docker://ghcr.io/ublue-os/aurora:stable
                   Digest: sha256:109a1480aefa1faef8f37b5de7ae143c2b2c961cd372c2415ec6fe7777944efe
                  Version: 40.20240807.0 (2024-08-07T13:32:18Z)
          LayeredPackages: corectrl distrobox docker java-17-openjdk kontact micro
```

ThoriumTextile:
Also `wine-core` should definitively be included by default too.

sekalengrengginang:
> Also `wine-core` should definitively be included by default too.
is there any use case for wine being included in base image? since for wine using Bottles/Lutris is preferred

RealVishy:
> Also `wine-core` should definitively be included by default too.

Is there any benefit to including it in the base image?I imagine most people interact with wine/proton through steam/lutris/bottles.

ThoriumTextile:
@RealVishy @sekalengrengginang Yes, because the native wine implementation works better.  Also there's no way to interact through CLI with flatpak Wine AFAIK.

sekalengrengginang:
[...]

PR #1603, opened 2024-08-14
feat: add breeze-gtk for aurora

RealVishy:
https://github.com/ublue-os/bluefin/issues/1596

Adds breeze-gtk package for more native theming of gtk apps on KDE Plasma.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -69,6 +69,7 @@
 				"zenity"
 			],
 			"kinoite": [
+				"breeze-gtk",
 				"skanpage",
 				"libadwaita-qt5",
 				"libadwaita-qt6",


castrojo:
Why are we adding themes we're trying to get rid of them?

RealVishy:
> Why are we adding themes we're trying to get rid of them?

I wasn't aware themes were being removed. This theme is mostly to improve the KDE experience with gtk apps by making them look more in line with qt apps.

castrojo:
Do we know why it's not shipped in Kinoite by default? `

castrojo:
(oops, misclicked, sorry), do we know why Kinoite doesn't ship this ootb?

RealVishy:
image

@castrojo I created a vm of kinoite in gnome boxes and it shows that breeze-gtk is installed by default.

m2Giles:
Let me check Aurora, but if this is in kinoite it should be in our image already.

PR #1605, opened 2024-08-14
fix: disable evdi

castrojo:

### DIFF
--- a/build_files/install-akmods.sh
+++ b/build_files/install-akmods.sh
@@ -27,12 +27,6 @@ rpm-ostree install \
     /tmp/akmods/kmods/*v4l2loopback*.rpm
     # /tmp/akmods-rpms/kmods/*framework-laptop*.rpm
 
-# All but Asus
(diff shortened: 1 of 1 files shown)


p5 (approved):

m2Giles:
Most of the back and forth you see is that the kmod breaks often. To not prevent builds we disable evdi.

Where things get tricky is that you cannot do an rpm-install for the kmod at run time. So if you want it you need to make a custom image. Displaylink is used in a ton of docks so we would like to include this for hardware enablement.

But yes overall it is a frustrating situation that this kmod breaks a lot. It has even more bizarre behavior as well.

castrojo:
Yeah sorry, not intentional on our part. Feels like we turn this thing on half of people's docks break, we turn it off the other half of people's docks break. And then the module may or may not build at that time. Additionally there's CPU usage issues depending on the dock being used, etc.   

I personally do not own one of these so we are dependent on what people report. It might make sense for us to just explicitly say we don't support these. I cannot believe vendors sell this as a product. :-/

PR #1630, opened 2024-08-27
feat: enable fsync kernels on the stable channel

castrojo:


castrojo:
This is punted for a while, closing.

PR #1656, opened 2024-09-08
fix:[#1651] Add MOTD message for SB keys

### LINKED ISSUE #1651: Add an motd warning when ublue's keys are not enrolled
We need to run this for a while in the MOTD and do a corresponding announcement on the forums. Let's block bringing fsync to stable. It'd help avoid situations like this: https://github.com/ublue-os/bluefin/issues/1636

```
WARNING: This machine has secure boot turned on but you haven't enrolled Universal Blue's keys, follow <link>
```

jardon:
- Add logic to check for SB enrollment and keys
- Update motd template

Closes #1651 

image

### DIFF
--- a/system_files/shared/usr/libexec/ublue-motd
+++ b/system_files/shared/usr/libexec/ublue-motd
@@ -23,6 +23,18 @@ if [[ -f "$TIP_FILE" ]]; then
 	fi
 
 	TIP_ESCAPED=$(escape "$TIP")
-
-	sed -e "s/%IMAGE_NAME%/$IMAGE_NAME_ESCAPED/g" -e "s/%IMAGE_TAG%/$IMAGE_TAG_ESCAPED/g" -e "s/%TIP%/$TIP_ESCAPED/g" /usr/share/ublue-os/motd/bluefin.md | tr '~' '\n' | /usr/bin/glow -s auto -w 78 -
 fi
+
+# check for secure boot key
+KEY_WARN=""
+FINGERPRINT="2B:E9:91:E3:B1:B5:40:70:F4:3D:80:BB:13:EB:C6:57:E5:A3:78:0D"
+mokutil --list-enrolled | grep -q $FINGERPRINT
+ENROLLED=$?
+mokutil --sb-state | grep -q enabled
+SB_ENABLED=$?
+
+if [[ $ENROLLED -eq 1 ]] && [[ $SB_ENABLED -eq 0 ]]; then
+	KEY_WARN="**WARNING**: This machine has secure boot turned on, but you haven't enrolled Universal Blue's keys. Failing to enroll these before rebooting **may cause your system to fail to boot**. Follow this [link](https://docs.projectbluefin.io/introduction#secure-boot) ~for instructions on how to enroll the keys."
+fi
+
+sed -e "s/%IMAGE_NAME%/$IMAGE_NAME_ESCAPED/g" -e "s/%IMAGE_TAG%/$IMAGE_TAG_ESCAPED/g" -e "s/%TIP%/$TIP_ESCAPED/g" -e "s|%KEY_WARN%|$KEY_WARN|g" /usr/share/ublue-os/motd/bluefin.md | tr '~' '\n' | /usr/bin/glow -s auto -w 78 -
--- a/system_files/silverblue/usr/share/ublue-os/motd/bluefin.md
+++ b/system_files/silverblue/usr/share/ublue-os/motd/bluefin.md
@@ -15,3 +15,5 @@
 - 󰈙 [Documentation](http://docs.projectbluefin.io/)
 - 󰊌 [Discuss](https://community.projectbluefin.io/)
 - 󰊌 [Leave Feedback](https://feedback.projectbluefin.io)
+
+    %KEY_WARN%
--- a/system_files/kinoite/usr/share/ublue-os/motd/bluefin.md
+++ b/system_files/kinoite/usr/share/ublue-os/motd/bluefin.md
@@ -15,3 +15,5 @@ Let's trace the stars.
 - 󰊤 [Issues](https://issues.projectbluefin.io)
 - 󰈙 [Documentation](http://docs.projectbluefin.io/)
 - 󰊌 [Discuss](https://community.projectbluefin.io/)
+
+%KEY_WARN%


KyleGospo (approved):

jardon:
These changes are untested.  I'll try to get an environment set up to test them before marking the PR as ready.

system_files/silverblue/usr/share/ublue-os/motd/bluefin.md
noelmiller, line 1: You will also need to add the message here too: https://github.com/ublue-os/bluefin/blob/main/system_files/kinoite/usr/share/ublue-os/motd/bluefin.md

noelmiller (approved):
LGTM, I will let others weigh in though.

KyleGospo (approved):

noelmiller:
> > The word wrapping is a little broken, but I believe this is an issue with glow and their handling of hyperlinks. Marking as ready since its out of my control and ugly warnings are better than no warnings at all.
> 
> I did a workaround for this, you can use `~` in your text and it will become a new line in the final output.
> 
> Good for another PR though, I don't want to hold this up over some formatting. LGTM.

Is this warning something we would need in Bazzite? Only case I can think of is if someone were to install Bazzite without secure boot on and then turn it on? But then they wouldn't be able to see the message anyway?

KyleGospo:
Missed this, seems there's a typo here? 
``if [[ $ENROLL -eq 1 ]] && [[ $SB_ENABLED -eq 0 ]]; then```

$ENROLL doesn't exist, $ENROLLED is created above.

jardon:
I'm not sure how I did that, but its fixed.  Should be good after the CI clears.  I rechecked it in my environment.

noelmiller (approved):

castrojo:
Let's merge this tomorrow (after the :stable builds go out), that way we have a week of it in :latest if we need to tweak the text.

jardon:
if [this](https://github.com/ublue-os/bluefin/pull/1661) merges first, i can refactor this PR to consolidate the logic

bsherman (approved):

system_files/shared/usr/libexec/ublue-motd
```diff
@@ -23,6 +23,18 @@ if [[ -f "$TIP_FILE" ]]; then
 	fi
 
 	TIP_ESCAPED=$(escape "$TIP")
+fi
+
+# check for secure boot key
+KEY_WARN=""
+FINGERPRINT="2B:E9:91:E3:B1:B5:40:70:F4:3D:80:BB:13:EB:C6:57:E5:A3:78:0D"
+mokutil --list-enrolled | grep -q $FINGERPRINT
```
bsherman, line 31: I don't think this actually works.

Example:
```
$ FINGERPRINT="2B:E9:91:E3:B1:B5:40:70:F4:3D:80:BB:13:EB:C6:57:E5:A3:78:0D"
$ mokutil --list-enrolled |grep $FINGERPRINT
```

Notice there is no output because the grep match failed...

I think you want something like this:
```
$ mokutil --list-enrolled --verbose-listing|grep -i $FINGERPRINT
SHA1 Fingerprint: 2b:e9:91:e3:b1:b5:40:70:f4:3d:80:bb:13:eb:c6:57:e5:a3:78:0d
```

bsherman (changes requested):
Correct me if I'm wrong, but I believe this needs to be fixed.

system_files/shared/usr/libexec/ublue-motd
```diff
@@ -23,6 +23,18 @@ if [[ -f "$TIP_FILE" ]]; then
 	fi
 
 	TIP_ESCAPED=$(escape "$TIP")
+fi
+
+# check for secure boot key
+KEY_WARN=""
+FINGERPRINT="2B:E9:91:E3:B1:B5:40:70:F4:3D:80:BB:13:EB:C6:57:E5:A3:78:0D"
```
bsherman, line 30: I'm not sure if we want to hard code the fingerprint or read it from the cert on-image?

```
openssl x509 -fingerprint -noout -in /etc/pki/akmods/certs/akmods-ublue.der
```

system_files/shared/usr/libexec/ublue-motd
jardon: it does work, but not by comparing output.  when grep fails to find a match, it gives an exit code of 1.  so this just compares the exit code to see if it matches or not.

I've tested it in a VM and it works well.  if there's a better option, I'm open to changing it

bsherman: I understand the `grep -q` usage, my point is it will always think the key is not enrolled because of the output from `mokutil --list-enrolled` vs `mokutil --list-enrolled --verbose-listing` 

```
$ mokutil --list-enrolled 
2bb010e24d fedoraca
2be991e3b1 ublue kernel
```
The output above will never have the fingerprint contents. 

Adding `--verbose-listing` does add the fingerprint output to the listing, however, with a lowercased fingerprint, which won't match what you have here.

system_files/shared/usr/libexec/ublue-motd
jardon: this is a good suggestion.  ive implemented this in the notify-send PR

system_files/shared/usr/libexec/ublue-motd
jardon: thats not what the included mokutil command produces
```
mokutil --list-enrolled | grep Finger
SHA1 Fingerprint: 2b:b0:10:e2:4d:94:c6:32:24:58:89:ba:aa:9e:d0:f3:d5:ef:1f:68
SHA1 Fingerprint: 2b:e9:91:e3:b1:b5:40:70:f4:3d:80:bb:13:eb:c6:57:e5:a3:78:0d
```

bsherman:
```
$ which mokutil --version
GNU which v2.21, Copyright (C) 1999 - 2015 Carlo Wood.
GNU which comes with ABSOLUTELY NO WARRANTY;
This program is free software; your freedom to use, change
and distribute this program is protected by the GPL.


$ /usr/bin/mokutil --list-enrolled | grep Finger

```

Does Bluefin somehow modify mokutil default behavior?

bsherman: image

bsherman: @jardon and I discovered in a side chat that the difference here is Fedora 39 has mokutil version 0.6.0 and Fedora 40 has version 0.7.1. 

And the outputs differ.

jardon:
closing in favor of https://github.com/ublue-os/bluefin/pull/1661

PR #1661, opened 2024-09-09
feat: Add notification for secure boot key check

### LINKED ISSUE #1651: Add an motd warning when ublue's keys are not enrolled

jardon:
- Add script that checks secure boot config and handles motd and GUI notifications
- Add systemd service to run script
- Enable systemd service
- Add motd key warning section

Closes #1651 

image
image
```
○ check-sb-key.service - Service to check for secure boot key enrollment
     Loaded: loaded (/usr/lib/systemd/system/check-sb-key.service; enabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: inactive (dead) since Thu 2024-09-19 06:54:01 CDT; 2s ago
   Duration: 67ms
    Process: 6815 ExecStart=/usr/libexec/check-sb-key.sh (code=exited, status=0/SUCCESS)
   Main PID: 6815 (code=exited, status=0/SUCCESS)
        CPU: 11ms

Sep 19 06:54:01 fedora systemd[1]: Started check-sb-key.service - Service to check for secure boot key enrollment.
Sep 19 06:54:01 fedora check-sb-key.sh[6829]: /etc/pki/akmods/certs/akmods-ublue.der is not enrolled
Sep 19 06:54:01 fedora systemd[1]: check-sb-key.service: Deactivated successfully.
```


bsherman:
@jardon a comment I make to most contributors... once you start getting feedback on PRs, please don't force push. Doing so erases history of comments and their relevance which makes it harder to track the history of change.

We squash merge most/all PRs in the ublue-os repos, so cleaning up commits is not a concern.

Thanks!

system_files/shared/usr/bin/check-sb-key
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+FINGERPRINT="$(openssl x509 -fingerprint -noout -in /etc/pki/akmods/certs/akmods-ublue.der | cut -c 18-)"
+mokutil --list-enrolled | grep -q $FINGERPRINT
```
bsherman, line 4: This doesn't work as expected. The current command will always report that the key is not enrolled.

This is due to the output from `mokutil --list-enrolled` vs `mokutil --list-enrolled --verbose-listing`:
```
$ mokutil --list-enrolled 
2bb010e24d fedoraca
2be991e3b1 ublue kernel
```
The output above will never have the fingerprint contents.

Adding `--verbose-listing` adds the fingerprint output to the listing, however, with a lowercased fingerprint, which won't still won't match what you have here.
```
$ mokutil --list-enrolled --verbose-listing | grep -i sha1
SHA1 Fingerprint: 2b:b0:10:e2:4d:94:c6:32:24:58:89:ba:aa:9e:d0:f3:d5:ef:1f:68
SHA1 Fingerprint: 2b:e9:91:e3:b1:b5:40:70:f4:3d:80:bb:13:eb:c6:57:e5:a3:78:0d
```


Example of a working match:
```
$ mokutil --list-enrolled --verbose-listing | grep -i $FINGERPRINT
SHA1 Fingerprint: 2b:e9:91:e3:b1:b5:40:70:f4:3d:80:bb:13:eb:c6:57:e5:a3:78:0d
```

Which, converted to work in the script would be:
```
mokutil --list-enrolled --verbose-listing | grep -q -i $FINGERPRINT
```

bsherman (changes requested):

jardon:
Marking this as ready to review BUT there are two caveats:
1. I'm not sure what the code analysis wants me to double quote
2. I haven't confirmed the exit code for the `mokutil --check-key` on a system where the key is not imported.  (I tried to delete in my VM, but it wasn't clearing it even though it was queued and rebooted)

jardon:
Also, can we confirm the key location for a fedora 40 based image?

bsherman:
> Also, can we confirm the key location for a fedora 40 based image?

The SecureBoot cert location we care about for ublue is the same on 39 and 40:  `/etc/pki/akmods/certs/akmods-ublue.der`

jardon:
@bsherman can you help me with the following:
1. I need to confirm the error code for `mokutil --sb-state` on a system with secure boot enabled
2. I need to change the warning location to a location that is writable.  I don't have a working ublue system right now.  Whats a good writable location to creat the markdown file for the motd banner?

bsherman:
> @bsherman can you help me with the following:
You bet @jardon 

>     1. I need to confirm the error code for `mokutil --sb-state` on a system with secure boot enabled

```
[root@fedora ~]# mokutil --sb-state
SecureBoot enabled
[root@fedora ~]# echo $?
0
```
but also...
```
[root@fedora ~]# mokutil --sb-state
SecureBoot disabled
Platform is in Setup Mode
[root@fedora ~]# echo $?
0
```

So you can't use the error code to know the state, but you can grep the output.

>     2. I need to change the warning location to a location that is writable.  I don't have a working ublue system right now.  Whats a good writable location to creat the markdown file for the motd banner?

I might do something like: `/run/user-motd-sbkey-warn.md` ... so it's temporary...

system_files/shared/usr/libexec/sb-key-notify.sh
```diff
@@ -0,0 +1,36 @@
...
+  exit 1
+fi
+
+WARNING_MSG="This machine has secure boot turned on, but you haven't enrolled Universal Blue's keys. Failing to enroll these before rebooting **may cause your system to fail to boot**. Follow this link https://docs.projectbluefin.io/introduction#secure-boot ~for instructions on how to enroll the keys."
+KEY_WARN_FILE="/usr/share/ublue-os/motd/key-warning.md"
+
+mokutil --sb-state | grep -q enabled
+SB_ENABLED=$?
+
+mokutil --test-key /etc/pki/akmods/certs/akmods-ublue.der
```
bsherman, line 14: I suggest only running this test after confirming that secure boot is enabled

KyleGospo (approved):

KyleGospo (approved):

castrojo:
Ok we good on this one? Last call!

bsherman:
> Ok we good on this one? Last call!

i'm actively testing, but i don't think this service works as expected.

system_files/shared/usr/libexec/sb-key-notify.sh
```diff
@@ -0,0 +1,40 @@
...
+    exit 0
+fi
+
+if mokutil --test-key "$KEY_DER_FILE"; then 
+    if loginctl --help | grep -q "json=MODE"; then
+        JSON_ARG="--json=short"
+    fi
+    USER_ID=$(loginctl list-users --output=json ${JSON_ARG:+$JSON_ARG} | jq -r '.[] | .user')
+    XDG_DIR=$(loginctl show-user "$USER_ID" | grep RuntimePath | cut -c 13-)
+    sudo -u "$USER_ID" \
```
bsherman, line 26: This `sudo notify-send` doesn't work when run from a root systemd service as setup here.

You can test this in a VM by writing the script to `/usr/local/sbin/sb-key-notify.sh` and the service to `/etc/systemd/system/sb-key-notify.service` and make sure you change the path for the script in the service file..

Then, `systemctl daemon-reload` and `systemctl enable sb-key-notify` ... reboot and/or simply `systemctl start sb-key-notify`... nothing happens.

I haven't dug deep into why this fails, but there's a more direct approach which is more what I meant to suggest when I suggested a service. 

What I'd do is keep all the testing for if secure boot is enabled and if the key is enrolled or not in a script like this... but the script would be `sbkey-missing-check.sh` or something... and if the conditions warrant a notification, write a file to `/run/sbkey-missing-notify`.

Then we need this `notify-send` command NOT with the sudo, to run as the user, and that should get setup with a profile.d/skel combo of script and .desktop file, similar to how we do for the `bluefin-firstboot` feature.

jardon: I'm curious why it doesnt work for you.  I've tested it in my environment and it works great.  Could you check the output of the service?

I had things more split up beforehand.  I'm open to seeing what that looks like to implement it using a desktop file, but I'm a little puzzled how we got here.

jardon: more specificially:
```bash
bash -x /path/to/script
```

system_files/shared/usr/lib/systemd/system/sb-key-notify.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Service to check for secure boot key enrollment and send notifications
+
+[Service]
+ExecStart=/usr/libexec/sb-key-notify.sh
+
+[Install]
+WantedBy=multi-user.target
+
+[Timer]
```
bsherman, line 10: `[Timer]` is invalid in a service unit.

You'll need to create a separate timer unit with the Timer settings which references this service if you want it to run on a timer, but I don't know why this needs to run on a timer. Once at boot to find the state should be enough.

system_files/shared/usr/libexec/sb-key-notify.sh
bsherman: > I'm curious why it doesnt work for you. I've tested it in my environment and it works great. Could you check the output of the service?

I set this up again on Bluefin:GTS (like i referenced in my comment on the Timer section). My sb-key-notify.service is installed in /etc and the sb-key-notify.sh is in /usr/local/sbin 

I _did_ see the notification popup when running `systemctl start sb-key-notify` on this setup, however... upon a reboot.

This is the output from the service:
```
 journalctl --boot -u sb-key-notify
Sep 15 13:07:53 fedora systemd[1]: Started sb-key-notify.service - Service to check for secure boot key enrollment and send notifications.
Sep 15 13:07:53 fedora sb-key-notify.sh[2952]: /etc/pki/akmods/certs/akmods-ublue.der is not enrolled
Sep 15 13:07:53 fedora loginctl[3023]: Failed to look up user : No such process
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]: usage: sudo -h | -K | -k | -V
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]: usage: sudo -v [-ABkNnS] [-g group] [-h host] [-p prompt] [-u user]
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]: usage: sudo -l [-ABkNnS] [-g group] [-h host] [-p prompt] [-U user]
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]:             [-u user] [command [arg ...]]
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]: usage: sudo [-ABbEHkNnPS] [-r role] [-t type] [-C num] [-D directory]
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]:             [-g group] [-h host] [-p prompt] [-R directory] [-T timeout]
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]:             [-u user] [VAR=value] [-i | -s] [command [arg ...]]
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]: usage: sudo -e [-ABkNnS] [-r role] [-t type] [-C num] [-D directory]
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]:             [-g group] [-h host] [-p prompt] [-R directory] [-T timeout]
Sep 15 13:07:53 fedora sb-key-notify.sh[3051]:             [-u user] file ...
Sep 15 13:07:53 fedora systemd[1]: sb-key-notify.service: Deactivated successfully.
```

bsherman: more context:
```
❯ systemctl cat sb-key-notify.service 
# /etc/systemd/system/sb-key-notify.service
[Unit]
Description=Service to check for secure boot key enrollment and send notifications

[Service]
ExecStart=/usr/local/sbin/sb-key-notify.sh

[Install]
WantedBy=multi-user.target

[Timer]
OnBootSec=1min
OnUnitActiveSec=3h

# /usr/lib/systemd/system/service.d/10-timeout-abort.conf
# This file is part of the systemd package.
# See https://fedoraproject.org/wiki/Changes/Shorter_Shutdown_Timer.
#
# To facilitate debugging when a service fails to stop cleanly,
# TimeoutStopFailureMode=abort is set to "crash" services that fail to stop in
# the time allotted. This will cause the service to be terminated with SIGABRT
# and a coredump to be generated.
#
# To undo this configuration change, create a mask file:
#   sudo mkdir -p /etc/systemd/system/service.d
#   sudo ln -sv /dev/null /etc/systemd/system/service.d/10-timeout-abort.conf

[Service]
TimeoutStopFailureMode=abort
```

jardon: can you post the output of `bash -x /usr/local/sbin/sb-key-notify.sh`, please?

system_files/shared/usr/lib/systemd/system/sb-key-notify.service
jardon: ya youre right about the timer.  i can edit it to run at first boot instead

bsherman: > ya youre right about the timer. i can edit it to run at first boot instead

it already runs at first boot.

system_files/shared/usr/libexec/sb-key-notify.sh
bsherman: > can you post the output of `bash -x /usr/local/sbin/sb-key-notify.sh`, please?

I can, and I will, but I've told you your script works, I do get a notification if running `systemctl start sb-key-notify` while already logged in.

The problem is notification from the script upon a boot. I'm not sure if you missed where i showed journal output with the failure.

bsherman:
```
root in ~ 
❯ bash -x /usr/local/sbin/sb-key-notify.sh
++ id -u
+ '[' 0 -ne 0 ']'
+ WARNING_MSG='This machine has secure boot turned on, but you haven'\''t enrolled Universal Blue'\''s keys. Failing to enroll these before rebooting **may cause your system to fail to boot**. Follow this link https://docs.projectbluefin.io/introduction#secure-boot ~for instructions on how to enroll the keys.'
+ KEY_WARN_FILE=/run/user-motd-sbkey-warn.md
+ KEY_DER_FILE=/etc/pki/akmods/certs/akmods-ublue.der
+ mokutil --sb-state
+ grep -q enabled
+ SB_ENABLED=0
+ '[' 0 -ne 0 ']'
+ mokutil --test-key /etc/pki/akmods/certs/akmods-ublue.der
/etc/pki/akmods/certs/akmods-ublue.der is not enrolled
+ loginctl --help
+ grep -q json=MODE
++ loginctl list-users --output=json
++ jq -r '.[] | .user'
+ USER_ID=bsherman
++ loginctl show-user bsherman
++ grep RuntimePath
++ cut -c 13-
+ XDG_DIR=/run/user/1000
++ echo 'This machine has secure boot turned on, but you haven'\''t enrolled Universal Blue'\''s keys. Failing to enroll these before rebooting **may cause your system to fail to boot**. Follow this link https://docs.projectbluefin.io/introduction#secure-boot ~for instructions on how to enroll the keys.'
++ tr -d '*~'
+ sudo -u bsherman DISPLAY=:0 DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus notify-send WARNING 'This machine has secure boot turned on, but you haven'\''t enrolled Universal Blue'\''s keys. Failing to enroll these before rebooting may cause your system to fail to boot. Follow this link https://docs.projectbluefin.io/introduction#secure-boot for instructions on how to enroll the keys.' -i dialog-warning -u critical -a mokutil --wait
+ echo '**WARNING**: This machine has secure boot turned on, but you haven'\''t enrolled Universal Blue'\''s keys. Failing to enroll these before rebooting **may cause your system to fail to boot**. Follow this link https://docs.projectbluefin.io/introduction#secure-boot ~for instructions on how to enroll the keys.'
```

jardon: oh i misunderstood.  ya so on boot, there is no user logged in so loginctl doesn't list a user to pass to the sudo command.  ill look into another solution then

bsherman: > oh i misunderstood. ya so on boot, there is no user logged in so loginctl doesn't list a user to pass to the sudo command. ill look into another solution then

yes, sorry it wasn't more clear.

bsherman: > oh i misunderstood. ya so on boot, there is no user logged in so loginctl doesn't list a user to pass to the sudo command. ill look into another solution then

This is why I proposed writing a state file from the service, and then having the user's login process (eg, autostart .desktop file) look for that file and run the notification. Then it would work for ANY user logged in.

bsherman:
> If anyone wants to test this in a secure boot enabled environment, rebase onto `ghcr.io/jardon/bluefin:key-dbus-notify`

Thank you for sticking with this PR! But also, thanks for going the extra mile and giving us a rebasable image for testing; that's excellent!

system_files/shared/usr/bin/sb-key-notify
```diff
@@ -0,0 +1,9 @@
+KEY_WARN_FILE="/run/user-motd-sbkey-warn.md"
+
+if [ -e $KEY_WARN_FILE ]; then
+    WARNING_MSG=$(cat "$KEY_WARN_FILE")
+
+    notify-send "WARNING" "$(echo "$WARNING_MSG" | tr -d '*~')" -i dialog-warning -u critical -a mokutil --wait
+else
+    echo "No secure boot key warning file.  Skipping...
```
bsherman, line 8: Missing a closing quote after `Skipping...`

Results in this error when the script runs and they key is not enrolled:

```
❯ /usr/bin/sb-key-notify 
/usr/bin/sb-key-notify: line 8: unexpected EOF while looking for matching `"'
```

bsherman (changes requested):
image

One change to make... a missing double quote in the `sb-key-notify` but otherwise, i think this is working.

I tested using the image you provided. Thank you! That made it much easier!

And apologies for taking so long to get back to it.

bsherman (approved):

bsherman:
> I've updated the test image

Uh, any idea why this link is duplicated?

image

jardon:
> > I've updated the test image
> 
> @jardon any idea why this link is duplicated?
> 
> image

upstream bug found here: https://github.com/charmbracelet/glamour/issues/347

i could make it a non-plain link but it still prints out the link so it would look more like `link-name link-address` instead of `link-address link-address`

bsherman:
> upstream bug found here: [charmbracelet/glamour#347](https://github.com/charmbracelet/glamour/issues/347)
> 
> i could make it a non-plain link but it still prints out the link so it would look more like `link-name link-address` instead of `link-address link-address`

No worries, sounds like something which will eventually get fixed upstream.

Still good with me.

@castrojo @KyleGospo I'm good with approval/merge on this now.

PR #1674, opened 2024-09-14
feat: add active directory support

castrojo:
We were missing a bunch of packages for this: https://github.com/fedora-silverblue/issue-tracker/issues/550

Gives us feature parity with Workstation. When we switch to bootc we should just pull in `domain-client` instead.


p5 (approved):
3MB is nothing, so LGTM

woprandi:
I just tried bluefin for the first time today and still not have `freeipa-client` installed

Issue #1672, opened 2024-09-14
Things to steal from Bazzite


castrojo:
This is a tracking issue. 

We need to catch up a few things from bazzite. I'll make a checklist and then later file these as individual bugs (or if someone wants to do that I can link them, it'd help me out since I'm so busy!)

- [ ] Link for Discourse shipped in the menu with the nice colored ublue logo. I think this is just copying a .desktop file and an icon.
- [ ] ~~Ship Just Perfection gnome extension - ok, the "Faster" setting for animations is awesome, we could try to figure out the actual value with our existing method but as I looked around in the tool I can see why people like it. Let's just put it in there and that's one less thing for people to complain about GNOME. Bazzite is large enough that I don't worry about extension porting and maintenance like I used to.~~

EDIT: We updated the animation speed to be faster, that's fine, users can install just perfection if they want.

castrojo:
Thanks for your help on this one!

PR #1693, opened 2024-09-24
feat: add Discourse desktop shortcut

egourves:
Add Discourse desktop shortcut.

Related to #1672 .

Based on [Discourse logo](https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/share/ublue-os/bazzite/discourse.svg) and [Discourse desktopfile](https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/share/applications/discourse.desktop) from Bazzite.


castrojo:
Awesome, thanks! We should put this and the documentation one in their own folder, but that can come later!

PR #1697, opened 2024-09-26
fix(ci): update rechunk action inputs

p5:
Have replicated the approach (i.e. rechunking rootless podman image), and it's working.


p5:
Let's hold on this PR for another hour or two.
I very much want to use rootless builds so we can disable and enable rechunk in a single line, but don't want to risk pushing a build that doesn't allow updates.

PR #1709, opened 2024-09-28
feat: fix weird titlebar on chromium

dreamyukii:
will fix this [issue on Chromium based browser ](https://github.com/flathub/org.gtk.Gtk3theme.Breeze/issues/94)


castrojo:
Thanks!

PR #1716, opened 2024-09-28
chore: Switch from dconf to gschema-overrides

fiftydinar:
# What I did

I removed `gdm` profile because it's already present in `/usr/share/dconf/profile/`.

I removed touchpad tap-to-click setting, because it's already the default since Gnome 46 (next GTS will be on Gnome 46).

I removed `user` profile because it's already present in `/etc/dconf/profile/`.

I moved dconfs from `local.d` to `distro.d`, as that's recommended to do as a distribution/custom image maintainer. `local.d` is intended for local-users, so let's make that easier for them.

Relocatable schemas & locked settings are not supported in gschema-overrides, hence why they remained in dconf.

I commented the gschema override, to make it easier to see where relocatable schemas are located.

`gsettings list-relocatable-schemas` can be used to check those on booted system.

We use `zz0-` prefix for gschema override, so that we can assure that Fedora gschema-override defaults won't override Bluefins.  Downstreams can use `zz1-`, `zz2-`, `zz3-`, etc. prefix, like I recommend BlueBuild users to do so.

Gschema overrides can be checked in build-time for errors in case some keys get migrated or someone makes an unintentional typing mistake, where dconf ones can't. So let's utilize that to our advantage.

# Testing

Bluefin should work without problems, but I'll have to additionally test Aurora, to see how Ptyxis dconfs are handled there.


fiftydinar:
Tested Aurora, it works great

fiftydinar:
Tested & fixed remaining Bluefin settings, everything works!

This is ready to merge

castrojo:
> I removed touchpad tap-to-click setting, because it's already the default since Gnome 46 (next GTS will be on Gnome 46).

What happens to existing gts users if this merges right now?

fiftydinar:
> What happens to existing gts users if this merges right now?

They'll loose this setting by default

castrojo:
Ok what about leaving it in for now but leaving a comment like `# TODO: Remove this after F40 becomes gts` and then we can clean it up after?

fiftydinar:
> Ok what about leaving it in for now but leaving a comment like `# TODO: Remove this after F40 becomes gts` and then we can clean it up after?

No problem, I'll add it back

fiftydinar:
Added

bsherman (dismissed):
I did a local image build of bluefin (not aurora) with these changes. Nothing sticks out to me as problematic.

And as I mentioned in Discord, I'm using the same general approach in my custom image, too. So this seems like a fair change to me.

fiftydinar:
@bsherman There is this PR, which modifies dconf.
https://github.com/ublue-os/bluefin/pull/1729

What do you think about merging that, then I will update this PR with those changes, @castrojo approves & merge?

castrojo:
Great idea, I merged that PR, ready to go!

fiftydinar:
@castrojo Ready!

PR #1719, opened 2024-09-28
fix: move to packaged topgrade

### LINKED ISSUE #1718: Newer topgrade package
### Describe the package

I see that you install `topgrade` from [pip](https://github.com/ublue-os/bluefin/blob/7b8091e7e882f21f8d1a9ad2847c5633a696cf48/build_files/fetch-install.sh#L16). The version provided by pip is 14.0.1 whilst the latest `topgrade` release is [15.0.0](https://github.com/topgrade-rs/topgrade/releases/tag/v15.0.0). It would be nice to get the latest and greatest as it includes a nice few things like [mise support](https://github.com/topgrade-rs/topgrade/pull/757)

### Image

All Images

castrojo:
Fixes #1718 

Bazzite has this in staging already so this should align us with them instead of installing from pip. 

We should test this though, I remember the upgrade to 15 causing confusion with some users, though we generally scope it down more than that compared to bazzite.

### DIFF
--- a/build_files/fetch-install.sh
+++ b/build_files/fetch-install.sh
@@ -12,9 +12,6 @@ echo 'eval "$(starship init bash)"' >> /etc/bashrc
 # Bash Prexec
 curl -Lo /usr/share/bash-prexec https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh
 
-# Topgrade Install
-pip install --prefix=/usr topgrade
-
 # Install ublue-update -- breaks with packages.json disable staging to use bling.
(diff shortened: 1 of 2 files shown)


p5 (approved):

castrojo:
Ok so tried this locally and it asks the user to accept changes in the git plugin with a confirmation prompt. Looks fine to me, scrolling back through some user chat I think new users get confused on what to put in there, which we can document.

castrojo:
image

KyleGospo:
There's an environment variable to skip that which must be set in ublue update and can optionally be set globally, though I would recommend leaving it for interactive use cases.

Malix-Labs:
[Topgrade 16 has been released](https://github.com/topgrade-rs/topgrade/releases/tag/v16.0.0)

PR #1722, opened 2024-09-29
fix: switch latest images to use ungated upstream kernel

### LINKED ISSUE #1717: Switch latest back to vanilla Fedora kernel
Bazzite needs this more than we do and some kernel developers have advised me to just keep it all stock. 

sched ext is coming in 6.12 anyway, let's go ahead and commit to ungated upstream kernel on latest. No change to stable/gts since those are on gated fedora kernels already.

p5:
Closes #1717 

This PR switches Bluerora to use the kernel version provided by upstream (still through kernel-cache), as recommended by some kernel developers.


bsherman (approved):

castrojo (approved):

bsherman:
I like this approach since it installs our dual-signed version of the kernel even if the version is the same as upstream Fedora.

PR #1732, opened 2024-10-02
feat: add specific flatpaks to dx images

noelmiller:
Podman Desktop is an example app that we would want to add specifically only to DX images. We can add others over time


p5 (approved):
LGTM!

PR #1748, opened 2024-10-06
chore: update to handle the negativo17-only main image

bsherman:
Now that main and hwe images are only providing negativo17, not rpmfusion, the 2 rpmfusion dependent akmods need some extra support, but other areas using rpmfusion can be cleaned up.


bsherman:
Note that https://github.com/ublue-os/akmods/pull/245 may remove the need to specify the dependencies for the rpmfusion kmod deps, but I won't be sure until will get that merged. 

I do think the current PR is good as is regardless.

m2Giles (approved):

PR #1750, opened 2024-10-06
chore: attempt to fix rechunker complaint of multiple kernels

bsherman:
PR #1722 passed all build checks, but after merging, we saw that for latest, where we use the stock upstream kernel, but with our added signature, rechunker complains of multiple kernels in /usr/lib/modules

Also, the method which was being used to remove the old kernel packages does not run when building locally on a Fedora host. So trying a different approach.

See: https://github.com/ublue-os/bluefin/actions/runs/11205608165/job/31145333321#step:23:233

### DIFF
--- a/build_files/cache_kernel.sh
+++ b/build_files/cache_kernel.sh
@@ -3,8 +3,10 @@
 set -eoux pipefail
 
 if [[ -n "${NVIDIA_TYPE:-}" ]]; then
-    rpm-ostree override remove \
-        kernel kernel-core kernel-modules kernel-modules-core kernel-modules-extra
+    for pkg in kernel kernel-core kernel-modules kernel-modules-core kernel-modules-extra
+    do
+        rpm --erase $pkg --nodeps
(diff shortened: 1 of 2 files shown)


build_files/cache_kernel.sh
```diff
@@ -3,8 +3,10 @@
 set -eoux pipefail
 
 if [[ -n "${NVIDIA_TYPE:-}" ]]; then
```
m2Giles, line 5: This variable is only set for gts/stable.

bsherman: well, that's part of the problem then... it's getting set for latest with the last PR which got merged...

castrojo (approved):

build_files/cache_kernel.sh
bsherman: I think this is good. If not, well, lets do another PR. :-)

noelmiller (approved):

PR #1754, opened 2024-10-07
chore(ci): replace vanilla kernels not asus/surface

bsherman:
Fixes bug in #1752 which ignored the fact that asus/surface images in Bluefin are sourced from the HWE images, thus they already have their expected kernels.

I do think this end state is a bit nicer though as it is easier to grok that this script only works on specific akmods flavors rather than the nvidia_type which didn't seem to have a clear correlation to replacing the kernel.

I apologize for the noise and confusion on this topic tonight.

### DIFF
--- a/build_files/cache_kernel.sh
+++ b/build_files/cache_kernel.sh
@@ -2,12 +2,14 @@
 
 set -eoux pipefail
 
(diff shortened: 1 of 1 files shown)


KyleGospo (approved):

bsherman:
This was weirdly stuck in the merge queue, even after all the merge checks were green. So I removed and manually merged.

Issue #1755, opened 2024-10-07
`ujust devmode` tries to put the user in a group that doesn't exist yet


castrojo:
![Image](https://github.com/user-attachments/assets/e4dad7c3-2a92-477d-a02c-4defcc3634b9)

Just tries to add you to the dx-group when doing `ujust devmode` but this fails because you haven't rebooted yet. It should recommend to run that next step, not try to run it right away.

m2Giles:
We need to have it copy the groups from /usr/lib/groups to /etc/groups first then add to the groups.

/usr/lib/groups are part of the read/only image so cannot append group membership there

PR #1765, opened 2024-10-10
fix(dx,just): copy dx groups to /etc/group before usermod

migeorge:
I just installed bluefin-dx on my framework laptop for the first time and found that my user account was not in the expected groups for docker, incus-admin, libvirt, and lxd.

I tried running `$ ujust dx-group` manually with no change to my user groups after logout or restart.

I found the groups were only defined in `/usr/lib/group` which cannot be modified.

This PR copies the groups from `/usr/lib/group` to `/etc/group` before the `usermod` commands are run.

I believe this relates to #1755

### DIFF
--- a/just/bluefin-system.just
+++ b/just/bluefin-system.just
@@ -126,6 +126,7 @@ ptyxis-transparency opacity="0.95":
 
 # Configure docker,incus-admin,lxd,libvirt container manager permissions
 dx-group:
+    cat /usr/lib/group | grep 'docker:\|incus-admin:\|lxd:\|libvirt:' | sudo tee -a /etc/group >/dev/null
     sudo usermod -aG docker $USER
     sudo usermod -aG incus-admin $USER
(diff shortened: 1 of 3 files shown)


just/bluefin-system.just
```diff
@@ -126,6 +126,7 @@ ptyxis-transparency opacity="0.95":
 
 # Configure docker,incus-admin,lxd,libvirt container manager permissions
 dx-group:
+    cat /usr/lib/group | grep 'docker:\|incus-admin:\|lxd:\|libvirt:' | sudo tee -a /etc/group >/dev/null
```
m2Giles, line 129: Please check if the group exists in /etc/group first. This appends them everytime.

m2Giles (changes requested):

m2Giles:
Going with https://github.com/ublue-os/bluefin/pull/1767 that already has that conditional check. Thank you!

PR #1767, opened 2024-10-10
fix: just dx-group append group entry from /usr/lib/group

obispobruno:
just dx-group would not add group to user. with this pr it checks if the group exists in `/etc/group`, and if not it appends from /usr/lib/group


m2Giles (approved):
Thank you!

bsherman (approved):

bsherman:
Ok, i see the problem
`/tmp/nvidia-install.sh: line 54: NVIDIA_FLAVOR: unbound variable`

https://github.com/ublue-os/bluefin/actions/runs/11306755007/job/31447703956#step:21:897

PR #1771, opened 2024-10-12
feat: add cryfs to bluefin

Raindrac:
This commit makes [io.github.mpobaschnig.Vaults](https://flathub.org/apps/io.github.mpobaschnig.Vaults) functional. Although cryfs can be installed by the user through brew or a distrobox, Vaults won't detect it. It must be installed directly onto the system to be detected by Vaults. Of the four months I've used Bluefin, this is the only package I've been stuck with layering.

Why provide support for Vaults? KDE Plasma contains a built-in feature called vaults that allows for the creation of locally encrypted containers. This can offer a second layer of encryption beyond LUKS or allow you to create encrypted containers on an otherwise unencrypted file system. This feature was added in Plasma 5.11.

Vaults provides support for that feature on Gnome, helping bring feature parity between the two desktop environments.

This commit only adds cryfs to Bluefin because it's already included in KDE Plasma as a backend for its own vault feature. Yep, this means that Aurora already has it! No need to add it again.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -60,7 +60,7 @@
 				"zsh"
 			],
 			"silverblue": [
-				"simple-scan",
+				"cryfs",
 				"gnome-shell-extension-appindicator",
 				"gnome-shell-extension-blur-my-shell",
 				"gnome-shell-extension-caffeine",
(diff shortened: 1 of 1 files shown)


castrojo:
Fine with this but please alphabetize it in the list, thanks!

Raindrac:
There we go, done! Both the silverblue and kinoite sections were out of order, so I used [Text Pieces](https://flathub.org/apps/io.gitlab.liferooter.TextPieces) to sort them both.

castrojo (approved):

salim-b:
@Raindrac I just tried to use the Vaults Flatpak with [`bluefin-dx:gts`](https://github.com/ublue-os/bluefin/pkgs/container/bluefin-dx/versions?filters[version_type]=tagged) that seems to have picked up the `cryfs` addition:

```sh
❯ rpm -qa | grep cryfs
  1799:	cryfs-0.11.3-4.fc39.x86_64
```

But Vaults still says

> No backends available. Please install gocryptfs or CryFS on your system.

Any idea what's going wrong?

Raindrac:
@salim-b I managed to reproduce this by rebasing to `bluefin:gts`. It seems to only affect `gts`, not `stable` and `latest`, and so the bug will likely fix itself once Fedora updates to Fedora 40 in a couple weeks. Still, I might create an issue about this later.

Using cryfs in `bluefin:gts` outputs the following error:
```cryfs: error while loading shared libraries: libfmt.so.10: cannot open shared object file: No such file or directory```

So, something clearly went wrong with one of the libraries that it uses here. Not sure why yet. Gotta look into it.

Issue #1736, opened 2024-10-03
Enable a `stable-daily` channel


castrojo:
Benjamin and I love the coreos gated kernel but miss being up to date with everything else. But we want weekly builds for `bluefin:stable` to remain the default.

So we're thinking of an opt-in `bluefin:stable-daily` that's just built daily. Then when zstd:chunked lands we can take a look at the update cadence in general and make adjustments when necessary. Then we just add it to the selector widget in `ujust rebase-helper`. 

We could either do it in another workflow or do it logically in the existing one, whatever makes the most amount of sense. No ISOs.

m2Giles:
We can have this be built daily and then do a check for tagging with stable weekly or if rebuild.

castrojo:
Also might be a good time to just line up the weeklies with gts, I think it would be easier cognitive load if we knew all the weeklies fired off at once, merging things on Sundays and Mondays feel "weird" since gts and stable get skewed, announcements need to mention this, etc.

m2Giles:
> Also might be a good time to just line up the weeklies with gts, I think it would be easier cognitive load if we knew all the weeklies fired off at once, merging things on Sundays and Mondays feel "weird" since gts and stable get skewed, announcements need to mention this, etc. 

We should definitely line everything up now. I'm okay with Monday UTC so it builds Sunday night in the US

castrojo:
Ok once the open PR adding the new workflow lands I'll reline everything up in a fresh PR.

PR #1779, opened 2024-10-13
feat: enable stable-daily channel

befanyt:
I really liked the idea of having a daily build of the stable channel as of talked in https://github.com/ublue-os/bluefin/issues/1736. So I tried my best to implement this feature.

Changes made:

- Aurora & Bluefin stable workflow now scheduled to build everyday and let you specify a "weekly_tag_day".
- If the workflow is not run via schedule, tag with stable and stable-daily.
- If the workflow is run via schedule and matches the weekly_tag_day, tag with stable and stable-daily.
- If the workflow is run via schedule but not matches the weekly_tag_day, tag with stable-daily.

Hope this is the correct implementation, thanks for this awesome project.

### DIFF
--- a/.github/workflows/build-coreos-aurora.yml
+++ b/.github/workflows/build-coreos-aurora.yml
@@ -8,7 +8,7 @@ on:
       - '**.md'
       - 'system_files/silverblue/**'
   schedule:
-    - cron: '41 5 * * 2'  # 5:41 UTC every Tuesday
+    - cron: '41 5 * * *'  # 5:41 UTC everyday
   workflow_dispatch:
 
 jobs:
@@ -20,4 +20,5 @@ jobs:
       brand_name: aurora
       fedora_version: stable
       rechunk: true
+      weekly_tag_day: Tuesday
 
--- a/.github/workflows/build-coreos-bluefin.yml
+++ b/.github/workflows/build-coreos-bluefin.yml
@@ -8,7 +8,7 @@ on:
       - '**.md'
       - 'system_files/kinoite/**'
   schedule:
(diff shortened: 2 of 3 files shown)


bsherman (approved):
I think this looks good:
- scheduled daily build of "stable" will result in tags: `stable-daily` and `stable-daily-TIMESTAMP`
- scheduled weekly build of "stable" will result in tags: `stable-daily`,`stable-daily-TIMESTAMP`, `stable`,`stable-TIMESTAMP`
- manual build will result in tags: `stable-daily`,`stable-daily-TIMESTAMP`, `stable`,`stable-TIMESTAMP`

In all cases the "DEFAULT_TAG" is `stable-daily` which i think is correct.

castrojo:
This is awesome, thank you for this!

bketelsen:
oh sweeeeet!

m2Giles:
Thank you for adding this!

bsherman:
@castrojo per our real-time chat, i think we may want an improvement where we can do a manual build which only regenerates a "stable-daily" and not a "stable-weekly" but that would also require a workflow specifically for the semi-rare scenarios where you need to manually build a weekly... and that would be extra logic in the tags.

castrojo:
Don't sweat the ISOs we can go back and clean those up, it's not burning through money. Though nothing got pushed to the bucket afaict. 

(On stable-daily here too, nice!)

castrojo:
Ok so I totally love how this is working out. I keep my main PC on stable-daily because I gotta always have everything, but then keep the rest of my PCs on stable because that's also fine. 

It's like a little sports mode button. 😈

PR #1785, opened 2024-10-14
feat(ci): build manual stable-daily or weekly

befanyt:
Follow up on the PR https://github.com/ublue-os/bluefin/pull/1779 where the need to manually build stable and stable-daily versions was expressed.

Introduces the following workflows, which can only be triggered manually.

- Aurora Stable Daily - (builds aurora:stable-daily)
- Aurora Stable Weekly - (builds aurora:stable)
- Bluefin Stable Daily - (builds bluefin:stable-daily)
- Bluefin Stable Weekly - (builds bluefin:stable)

This also required a few changes to the reusable-build workflow. The Bluefin Stable and Aurora Stable workflows will still produce both versions when run manually, and may produce one or both when triggered by a schedule. 

P.S. I hope someone can address the ISO build part that it produces, as I definitely cannot. :sweat_smile:

### DIFF
--- a/.github/workflows/build-coreos-aurora-daily.yml
+++ b/.github/workflows/build-coreos-aurora-daily.yml
@@ -1,0 +1,16 @@
+name: Aurora Stable Daily
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: build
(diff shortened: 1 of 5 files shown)


castrojo:
Don't we still need this?

PR #1792, opened 2024-10-16
feat: new autumn wallpaper

Raindrac:
This commit contains a new autumn wallpaper commissioned by @castrojo. It also changes the default wallpaper to that new wallpaper and it creates a framework for easily dropping in new monthly wallpapers.

New monthly wallpapers can be added in by dropping them into `/system_files/silverblue/usr/share/backgrounds/bluefin` with the appropriate file name and then uncommenting the corresponding xml file located in
`/system_files/silverblue/usr/share/gnome-background-properties`. Nothing else needs to be done as everything else is prepped and ready to go.

The following attachment is a spreadsheet that goes into detail about when day/night transitions occur for every month and the reason behind those decisions.

[Bluefin Wallpaper Timings.ods](https://github.com/user-attachments/files/17389537/Bluefin.Wallpaper.Timings.ods)


Raindrac:
Removing the compression applied to the wallpapers. I'll submit it as a separate commit once this one is approved. 🙂

Raindrac:
Removing the compression applied to the wallpapers. I'll submit it as a separate PR once this one is approved. 🙂

PR #1794, opened 2024-10-16
feat: new autumn wallpaper

Raindrac:
This commit contains a new autumn wallpaper commissioned by @castrojo. It also changes the default wallpaper to that new wallpaper and it creates a framework for easily dropping in new monthly wallpapers.

New monthly wallpapers can be added in by dropping them into `/system_files/silverblue/usr/share/backgrounds/bluefin` with the appropriate file name and then uncommenting the corresponding xml file located in
`/system_files/silverblue/usr/share/gnome-background-properties`. Nothing else needs to be done as everything else is prepped and ready to go.

The pcolor of every new wallpaper was set to #000000 so that, if for whatever reason a wallpaper fails to load, the problem is more obvious. Existing wallpapers were changed to match.

The following attachment is a spreadsheet that goes into detail about when day/night transitions occur for every month and the reason behind those decisions.

[Bluefin Wallpaper Timings.ods](https://github.com/user-attachments/files/17389537/Bluefin.Wallpaper.Timings.ods)


castrojo:
Hah, awesome! Can we get this going on the testing branch for a day or so?

Raindrac:
That'd be fine, though do keep in mind that I've already tested it pretty extensively on some virtual machines. Might not be necessary? I'll leave it up to y'all!

castrojo:
Yeah I just wanna test once or twice in different time zones, I appreciate the level of science put into this!

castrojo (approved):
Weekly builds are already out the door, might as well go for it, I have time to test today

Raindrac:
There we go. That should ensure backwards compatibility for users on the spring, summer and winter wallpapers. I'm going to leave the November wallpaper as-is; we can skip it by not changing the default until December and then merging in @p5's commit on December 1st so that it'll automatically switch from October to December.

castrojo (approved):

PR #1796, opened 2024-10-16
chore: compress images

Raindrac:
Monthly wallpapers currently range in size from 10 MB to 30 MB. Once we have all 24 monthly wallpapers, they're going to take up around ~480 MB of the installation size. This is large enough that they'll take up around 10% of the installation size of Bluefin. That's hefty!

The file size could be reduced by applying lossy compression.

This commit applies 90% lossy compression to all of the monthly wallpapers. Here's a comparison.

Screenshot from 2024-10-15 22-21-26

Can you spot the difference? The one on the right, which is lossy and included in this commit, has a slightly smoother film grain. And this is with 200% zoom; at 100% zoom, it's practically unnoticeable.

This marginal decrease in quality reduces file size by **85% - 95%** and at the moment it already saves **177.6MB**. If you ask me, it's definitely worth it! The gains will only continue as we add new wallpapers.

I believe there's merit to uploading the lossless variants somewhere, so I propose creating a new repository to host the lossless wallpapers alongside this commit.


m2Giles:
I think we should try to do this with all artwork in the repo and make a new top-level directory for Art Assets that are uncompressed.

We can then have a ujust to download the images to /usr/local/share if having an easy to consume method makes sense.

Looking at the repo right now, the images are the single largest component.

castrojo:
Bazzite switched to jpegxl for their wallpapers and had substantial savings, we should just match them and the upstream gnome wallpapers switched to that too.

Raindrac:
I like the idea of using jxl and compressing other files. I'll look into it, and for now I'll convert this commit to a draft

Raindrac:
Alright, that should do! I solved a problem where the wallpapers would fail to load if they were applied before this change by using symlinks. Could probably remove 'em in a year or two, by then most users will have changed their wallpaper and their removal would affect next to no-one.

This should be ready to go.

Raindrac:
Hold up, I'm going to convert this to a draft once more as I consider splitting it into multiple commits. There's a lot going on here now.

PR #1810, opened 2024-10-19
feat: theming updates

p5:


packages.json
```diff
@@ -77,7 +78,6 @@
 				"nautilus-open-any-terminal",
 				"openssh-askpass",
 				"simple-scan",
-				"yaru-theme",
```
castrojo, line 80: Can we keep it on disk? I'd still be useful for those who prefer to keep bluefin how it used to be.

p5: Yeah, sure.  Will add it back.
FWIW, if you like the Files icon, we can use only Yaru's icons

packages.json
```diff
@@ -77,7 +78,6 @@
 				"nautilus-open-any-terminal",
 				"openssh-askpass",
 				"simple-scan",
```
p5, line 80:
```suggestion
				"simple-scan",
				"yaru-theme",
```

PR #1812, opened 2024-10-20
enhance: rebase_helper docs

Malix-Labs:
illustration

lead to https://docs.projectbluefin.io/administration#upgrades-and-throttle-settings


castrojo:
We just added this text, it's fine.

Malix-Labs:
Don't you think the link could be good to add too ?

Issue #1804, opened 2024-10-18
Simplify reusable build tagging section


castrojo:
We're enabling F41 so we don't have time to clean this up so filing it in case someone wants to flex their tagging skills. 

M2 isn't happy with this tag section, too nested, too unreadable. A refactor here would be appreciated!

https://github.com/ublue-os/bluefin/blob/71a923264e334612d3d90dff3e2b196f560eacbc/.github/workflows/reusable-build.yml#L244

tulilirockz:
#1910 should fix this

tulilirockz:
Got fixed with that PR! This should be good now

PR #1815, opened 2024-10-20
refactor(ci): stable-daily tagging

befanyt:
I have refactored the stable-daily tagging logic to make it much more readable. I hope M2 is happier with this code.
Reference: https://github.com/ublue-os/bluefin/issues/1804

Psuedo logic:
```
if triggered by schedule and its not weekly_tag_day; then
    tag with stable-daily
else
    tag with stable stable-daily
fi
```
In the code, I used the explicit term `stable` instead of `${FEDORA_VERSION}` to improve readability. I chose this because we are already in stable logic, so this would always be `stable`. If you prefer `${FEDORA_VERSION}` I can, of course, put that back in.

Manual triggered run will produce: stable, stable-daily
Schedule triggered run on weekly_tag_day: stable, stable-daily
Schedule triggered run on non weekly_tag_days: stable-daily

On top of this I made changes that will allow you to manually build stable or stable-daily only by pushing a button.
Introduces the following workflows.

- Aurora Stable Daily - (builds aurora:stable-daily)
- Aurora Stable Weekly - (builds aurora:stable)
- Bluefin Stable Daily - (builds bluefin:stable-daily)
- Bluefin Stable Weekly - (builds bluefin:stable)

To make this possible I updated the above pseudo logic like so:
```
if triggered by schedule and its not weekly_tag_day; then
    tag with stable-daily
else if build_stable_weekly is false; then 
    tag with stable-daily
else if build_stable_daily is false; then 
    tag with stable
else
    tag with stable stable-daily
fi
```

`build_stable_weekly` and `build_stable_daily` are both set to `true` by default inn the `reusable-build.yml` and are toggled via the new workflow files.

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -20,6 +20,16 @@ on:
         required: false
         type: string
         default: Tuesday
+      build_stable_daily:
+        description: "Build with 'stable-daily' tag"
+        required: false
+        type: boolean
+        default: true
+      build_stable_weekly:
+        description: "Build with 'stable' tag"
+        required: false
+        type: boolean
+        default: true
     outputs:
       images:
         description: "An array of images built and pushed to the registry"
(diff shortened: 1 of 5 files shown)


.github/workflows/reusable-build.yml
```diff
@@ -20,6 +20,16 @@ on:
         required: false
         type: string
         default: Tuesday
+      build_stable_daily:
```
m2Giles, line 23: This is fine for now, but it feels a bit like a belt and suspenders situation. We are being overly specific it feels like.

.github/workflows/reusable-build.yml
```diff
@@ -246,15 +256,18 @@ jobs:
               COMMIT_TAGS+=("${SHA_SHORT}")
           fi
 
+          TODAY="$(date +%A)"
           if [[ ${{ matrix.fedora_version }} == "stable" ]]; then
```
m2Giles, line 260: This is much more readable now.

That said I think we have room for improvements in the future.

m2Giles (approved):

bsherman (approved):

castrojo:
Builds are skewed, forcing this in. Thanks everyone!

PR #1823, opened 2024-10-21
feat: show full Bluefin logo during system startup

Raindrac:
This commit changes the logo for Bluefin displayed in Plymouth and GDM during system startup. Previously, the Plymouth logo didn't include the project title and it used an image of Raven rather than Bluefin, so I changed it to the full logo with Bluefin. As for the GDM logo, it was changed to a logo variant without the "Project" subtitle for a cleaner appearance.

The logo was also slightly enlarged for readability. As a result of this commit, Plymouth and GDM now show the same logo, offering a feeling of consistency as the logo remains the same throughout every step of the boot process.

Tested on a virtual machine!


castrojo:
The raven was picked on purpose since it's darker and we want to keep that more subtle. Agreed on the other changes though.

Raindrac:
Personally, I find it a little too subtle. It gets easily overshadowed by the giant motherboard logo above it, which in my case is "GIGABYTE: INSIST ON ULTRA DURABLE". Well... it's a little bit of that, and showing the mascot of Bluefin more closely matches my expectations. Feels strange to not show her during the beginning of a boot.

PR #1832, opened 2024-10-25
feat: do not attach modal dialog windows

salim-b:
Attaching modal dialogs [introduces serious usability problems for various apps](https://www.reddit.com/r/gnome/comments/1dbzni6/dear_gnome_devs_attach_modal_dialogs_should_be/), the reason being such apps not following GNOME designer's intentions IIUC. Anyways, I think such *separate-window* modal dialogs are mostly a non/pre-libadwaita design pattern, so modern GNOME apps won't be negatively affected—but legacy ones will (often) gain usability.


castrojo:
Neither upstream or ubuntu set this to false (they set leave it at true). People can set this one by hand if they prefer it.

Issue #1841, opened 2024-10-27
[Discussion] Refactor Image Build


m2Giles:
The current image build process enters into either one of two scripts inside of build_files. These scripts then call the other scripts in that directory to build the image.

Positives: Contributors can get a bite sized piece of how the image is built and with it inside of a bash script we do not have to several && or ; with \ to chain together commands inside the Containerfile.

Downsides; Because each of these scripts are bite sized, its hard to know where to modify items. There is no real naming scheme and no indication if certain scripts need to be run in a certain order (which there is).

Proposal:
This needs to be cleaned up. Each script must have a descriptive name along with a number specifying order in which they must be executed. -dx and base specific scripts should be merged, and internal control flow logic can be used to make sure the right code is executed. Small scripts that must be run after another another script should instead be added to the script that the code depends on.

The Containerfile and reusable-build.yml should also be refactored to remove as many unnecessary build arguments that are either duplicative or can be easily determined inside the container at build time. Across the board, an eye towards simplification is needed understanding that complexity will be grafted on over time.

Issue #1840, opened 2024-10-27
GNOME theme fonts and icons lost after reset


DarkGhostHunter:
### Describe the bug

Fonts and icons are reset to default (or whatever is it) after rebooting the system. No changes to these were made.

![Image](https://github.com/user-attachments/assets/84991514-4143-4790-8eb4-9c433f4b2aa2)

![Image](https://github.com/user-attachments/assets/0ee2cc79-2868-4a93-86d4-127b8a8217cd)

This also affects GNOME Greeter (where you login for the first time)

### What did you expect to happen?

To keep them as is.

### Output of `rpm-ostree status`

```
State: idle
AutomaticUpdates: stage; rpm-ostreed-automatic.timer: no runs since boot
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bluefin-dx:stable
                   Digest: sha256:75fe45926ca23fe63414bc42b22f40e650decc1eae8d6d4b621e7d8e6b3721d0
                  Version: 40.20241027.0 (2024-10-27T05:46:52Z)

  ostree-image-signed:docker://ghcr.io/ublue-os/bluefin-dx:stable
                   Digest: sha256:6268e5ef31ce6e6bed0b07f0c46e8fadccbfa5b7f7594008d623db1ae72376a6
                  Version: 40.20241019.0 (2024-10-20T22:26:08Z)
```

### Output of `groups`

`awesomedev wheel`

### Extra information or context

This happened to me also on `bluefin-dx:latest` this week, and I thought I botched my Linux installation somewhere. I can confirm this bug also landed on `bluefin-dx:stable`.

This could be related to #1811, where KDE also lost theming.

castrojo:
This is on purpose: https://github.com/ublue-os/bluefin/issues/1685?issue=ublue-os%7Cbluefin%7C1683

The announcement of the switch will go out on tuesday during the release.

DarkGhostHunter:
That explains the default font, but not the icon set.

castrojo:
We reverted the entire theme to stock, including the icons: https://github.com/ublue-os/bluefin/pull/1813/files

DarkGhostHunter:
I get it, you had to revert back to default since the icons don't support accent color on latest GNOME, am I right?

I was so accustomed to Ubuntu that I felt back in 2000 when I saw Inter on the greeter

PR #1867, opened 2024-11-01
refactor: Reorganize Build Process

m2Giles:
This significantly reworks the build system for Bluefin/Aurora.

Changes:
Consolidates Workflows
Removes Surface and Asus images in favor of a single HWE image
Rewrite of local dev Justfile
Reorganization and Consolidation of build files

Add new files

Closes: #1841


.github/workflows/build-images.yml
```diff
@@ -0,0 +1,22 @@
...
+    
+permissions:
+    contents: read
+    packages: write
+    id-token: write
+
+jobs:
+    build-image-gts:
+        uses: ./.github/workflows/build-image-gts.yml
+        secrets: inherit
```
p5, line 13:
```suggestion
    build-image-gts:
        uses: ./.github/workflows/build-image-gts.yml
        secrets: inherit
        strategy:
          matrix:
            base_name: [bluefin, bluefin-dx, aurora, aurora-dx]
```

.github/workflows/build-image-gts.yml
```diff
@@ -1,23 +1,21 @@
...
   workflow_dispatch:
+  workflow_call: 
 
 jobs:
-  build:
-    name: build
+  build-image-gts:
+    name: Build GTS Images
     uses: ./.github/workflows/reusable-build.yml
     secrets: inherit
```
p5, line 18:
```suggestion
  build-image-gts:
    name: Build GTS Images
    uses: ./.github/workflows/reusable-build.yml
    secrets: inherit
    strategy:
      matrix:
        image_flavor: [none, hwe, nvidia]
```

m2Giles:
#1841
#1840

m2Giles:
#1841
#1840

castrojo:
@befanyt got any opinions here? If we're going to build a new world we might as well get new eyes. 😄

m2Giles:
Changelogs script is in place. It needs to be wired up.

In testing I saw that GTS and Latest do not have dates associated with them. That is rectified in this PR.

m2Giles:
Need to handle how stable-daily vs stable changelogs work.

Right now the workflow is called stable and we use that for changelogs. But we only want stable to get tagged weekly.

We could upload an artifact as a trigger. Or have some other way passing the fact that two changelogs are needed from resuable-build

m2Giles:
Remaining:

Changelogs running for GTS and weekly stable only. If the workflow is GTS run thr changelog. If it's stable (not stable daily) make an output to tell to run changelogs.

Kernel Pinning. Have an input to reusable build for pinning a kernel.

m2Giles:
I believe this is now ready for review.

bsherman (commented):
Looking this over... this is a hugely valuable effort.

However, I think we should go one step further. We should be using the value of the just recipes in our github action workflow.

At the least, the `just build` recipe should be used in `reusable-build.yml` to build the images.
All the tagging logic probably makes sense to remain in the workflow since that's external to the container build itself. Similarly, the push operation doesn't need to be in just, and it ties closely to the tags.

We could possibly use `just build-rechunk` in the workflow as well... Looking for feedback from @m2Giles on that.

But the primary value from doing this is ability to build container images the same way both local and in github actions. That's huge!

Fixing a bug in the image build can be validated before pushing to a PR even.

m2Giles:
I think the only thing needed to change to make that a reality is to make sure build container and rechunk can run completely as root.

The only major concern is not having enough space on builds when copying to the rootful podman store.

bsherman:
> I think the only thing needed to change to make that a reality is to make sure build container and rechunk can run completely as root.

Ah, like an additional flag to the build recipe to run it as root? Yes, then it would match workflow behavior.

> The only major concern is not having enough space on builds when copying to the rootful podman store.

This should be addressed by running build as root, no?

m2Giles:
Yes. I've already put in a bunch of sudoif's so this should be just testing to make sure it works as intended.

I may add the secureboot check to the justfile.

bsherman:
After further review and discussion, while we DO want to make use of the new and improved just recipes within the github workflows, that effort can be iterated on. As is, this is already large and holding up some needed fixes.  So we should make an issue to reduce the workflow specific code by calling just instead.

I'm continuing to review the existing changes tonight. So far looking nice.

bsherman (dismissed):
I have spent a bunch of time testing the justfile locally, looking through the build script changes (made a few changes there myself to fix numbering of build scripts), looking through the workflow yaml (fixed some naming conventions), and looking at the GREEN successful builds validating the podman build args.

I think this is looking good.


A great improvement and we'll keep on making it better.

Thanks @m2Giles !

bsherman (approved):

PR #1870, opened 2024-11-03
feat?: Make Ublue-specific desktop icons a bit more consistent with each other

tulilirockz:
I believe this could be a nice PR since the ublue-specific icons are quite distinct, and this attempts to fix this by changing them a little to match each other better. Im not an artist so I just got the ublue-docs icon and modified a bit to look like the other icons


tulilirockz:
image
image

fiftydinar (approved):
I like this

castrojo:
These are beautiful!

castrojo (approved):

Issue #1786, opened 2024-10-14
F40 - ublue-user-setup service regenerates fonts every time user logs in


tulilirockz:
This was used due to some workaround in the F39 version and probably doesn't need to happen anymore, it would be preferable if this could be disabled by default on F40 due to it making the desktop stutter a bit on lower-end hardware when logging in

PR #1871, opened 2024-11-03
fix: remove redundant fc-cache run on ublue-user-update service

tulilirockz:
On F40 this is apparently not needed anymore, and since GTS is already on F40, all versions used by Bluefin/Aurora are already fixed upstream - Fixes: #1786


fiftydinar:
Maybe it would be good to just comment it instead of removing, as flatpak font cache issue might happen again in the future

Otherwise, approve

fiftydinar (approved):

castrojo:
I think we can safely just nuke the entire thing, it didn't really work right in F39 either.

castrojo (approved):

PR #1876, opened 2024-11-04
feat: ensure firefox (flatpak) has hardware decoding turned on by default

### LINKED ISSUE #1730: Ensure firefox has hardware decoding turned on
Did a new reinstall on a device and I had to manually set hw accel on in Firefox: 

Bazzite has this: https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/share/ublue-os/firefox-config/01-bazzite-global.js

I could have sworn we fixed this ages ago and was surprised when the laptop turned into a jet engine.

tulilirockz:
Got this straight from Bazzite, just renamed a few thingies to make it Bluefin-themed :p - Fixes #1730


bsherman:
I'm not sure about all these settings... I don't think they are required to enable hardware accelerated video decoding.

I don't have any of them and hw accel decode works well for me. (I happen to have been testing this topic lately).

system_files/silverblue/usr/share/ublue-os/firefox-config/03-bluefin-gnome.js
bsherman, line 1: Let's skip this file for now as I'd like to do small, targeted changes and at the least, the rounded corner setting has nothing to do with hw accerlation.

bsherman (changes requested):
I think we can go with this if it's only the `01-bluefin-global.js` for now.

I've been testing hardware acceleration a lot this past week. So while I do want us to default enable `media.ffmpeg.vaapi.enabled` and `gfx.webrender.all` is also probably fine, I'm hesitant on the other changes just yet.

system_files/silverblue/usr/share/ublue-os/firefox-config/03-bluefin-gnome.js
tulilirockz: True! Lemme just fix this thing then

bsherman (approved):
Thank you!

PR #1897, opened 2024-11-06
feat(just): add 1password ujust install command

wiltaylor:
Added a the 1password gui and cli installer to ujust. 

Note: This needs to be installed in the base image because 1password acts as a ssh agent and hooks into the os.


castrojo:
Sorry we don't recommend layering ootb, this is more appropriate for a custom image.

PR #1898, opened 2024-11-06
feat(just): add nix derminate systems installer

wiltaylor:
Added a ujust entry to install nix using the determinate system installer. https://github.com/DeterminateSystems/nix-installer which supports OSes built on top of silverblue.

### DIFF
--- a/just/bluefin-apps.just
+++ b/just/bluefin-apps.just
@@ -102,3 +102,8 @@ install-incus:
         echo "Run \"just devmode\" to turn on Developer mode."
         exit
     fi
+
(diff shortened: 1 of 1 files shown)


p5:
Unfortunately, we decided to not support Nix since it caused many issues.  
We used to have a script which looks exactly like this, though found Nix messes with SELinux and causes issues.  Or at least it did just over a year ago.  

We'll need to investigate and see if this is something we'd like to try supporting again.

castrojo:
Sorry, we don't support nix.

PR #1910, opened 2024-11-08
chore(ci): Consolidate workflow into just commands

m2Giles:


m2Giles:
Need to add the just install action.

m2Giles:
Building the image looks good now.

Need to test build-iso action.

m2Giles:
This is now ready for review

m2Giles:
Reran checks. Found one more scheduled instead of schedule.

Made sure it runs locally again.

bsherman (approved):
we won't know for sure if its perfect until we merge due to some last stage only on non-PR steps, but we have caught a few and in general this is looking great!

PR #1914, opened 2024-11-10
chore(just): clean up ujust recipes to fit newer style

### LINKED ISSUE #1805: `ujust switch-channel` alias is broken
This is probably something simple I'm missing.

https://github.com/ublue-os/bluefin/blob/71a923264e334612d3d90dff3e2b196f560eacbc/just/bluefin-system.just#L262

tulilirockz:
Should fix #1805

### DIFF
--- a/just/bluefin-system.just
+++ b/just/bluefin-system.just
@@ -77,7 +77,8 @@ devmode:
         fi
     fi
 
-alias gnome-vrr := toggle-gnome-vrr
+gnome-vrr:
+    @ujust toggle-gnome-vrr
 
 # Enable or Disable Gnome-VRR
 toggle-gnome-vrr:
@@ -253,8 +254,14 @@ install-system-flatpaks:
 configure-grub:
     @/usr/libexec/configure-grub.sh
 
-alias rollback-helper := rebase-helper
-alias switch-channel := rebase-helper
+switch-stream:
+    @ujust rebase-helper
+
+switch-streams:
+    @ujust rebase-helper
+
+rollback-helper:
+    @ujust rebase-helper
 
 # Rebase assistant
 rebase-helper:
--- a/just/bluefin-system.just
+++ b/just/bluefin-system.just
@@ -316,19 +316,19 @@ toggle-tailscale:
         TS_OPTION=$(Choose Enable Disable)
 
         # gnome-extensions is only available on Bluefin (Gnome)
-        if [ "$VARIANT" == "silverblue" ]; then
-          TAILSCALE_QS="$(gnome-extensions list | grep -q "tailscale@joaophi.github.com"; echo $?)"
-          if [ "$TAILSCALE_QS" == 0 ]; then
-            TAILSCALE_QS="Installed"
-          else
-            echo "The Tailscale QS extension for Gnome is not installed. Please install it and then run this script again."
-          fi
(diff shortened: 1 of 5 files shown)


m2Giles:
We should also use this as an opportunity to clean up stale ones.

just/bluefin-system.just
```diff
@@ -253,8 +254,14 @@ install-system-flatpaks:
 configure-grub:
     @/usr/libexec/configure-grub.sh
 
-alias rollback-helper := rebase-helper
-alias switch-channel := rebase-helper
+switch-channel:
```
castrojo, line 257:
```suggestion
switch-stream:
```

just/bluefin-system.just
```diff
@@ -253,8 +254,14 @@ install-system-flatpaks:
...
+  @ujust rebase-helper
+
+switch-channels:
```
castrojo, line 260:
```suggestion
switch-streams:
```

castrojo: coreos calls these streams

castrojo (changes requested):

m2Giles:
when working with the justfiles. Please be sure to run `just fix`. We run `just check` as part of the workflow to make sure we don't ship broken justfiles.

tulilirockz:
Should I also include a move to bootc on this PR? I am thinking about refactoring most of these scripts to not use rpm-ostree and instead use a wrapper over bootc that allows it to run rootless (ubootc or something)

tulilirockz:
The indentation on some places also seems a bit funky, ill be reformatting them to always fit 4 spaces indentation

castrojo:
> Should I also include a move to bootc on this PR? 

Follow on PRs are always better instead of conflating things. 

> a wrapper over bootc that allows it to run rootless (ubootc or something)

This is out of scope for us, we're trying to delete as much of our custom stuff as possible these days.

just/bluefin-system.just
```diff
@@ -253,20 +272,35 @@ install-system-flatpaks:
...
-alias rollback-helper := rebase-helper
-alias switch-channel := rebase-helper
+# Rebase assistant
+[group('System')]
+switch-stream:
+    @ujust rebase-helper
+
+# Rebase assistant
+[group('System')]
+switch-streams:
```
m2Giles, line 282: just alias this.

tulilirockz:
Then we could regroup stuff, honestly I feel like separating it in `apps`, `tools` and `system` doesnt seem like the best? `apps` and `tools` could be the same category (`tools`)

castrojo:
We need to keep jetbrains (people comment about how useful this is all the time), and the tailscale toggle was an ask in the survey. So let's keep that for people who use other VPNs. Not sure on invokeai so let's keep it for now. We can always do another pass.

castrojo:
OK one final pass from others would be appreciated on this one! (I know m2 wanted one more look).

tulilirockz:
Alright, so, after manually testing every single one of the affected recipes, they all seem to work fine, without any major changes. The only ones that I saw that were broken were the toggle-tailscale and setup-ollama ones - just reverted them. The tailscale one also didnt seem to toggle the extension on the `Silverblue` $VARIANT so I just made it uppercase :p

castrojo (approved):

Issue #1912, opened 2024-11-09
`ujust update` nevers updates npm and displays message `npm: IGNORED`


marzvrover:
### Describe the bug

`ujust update` always ignores npm.

```shell
── 00:36:36 - Node Package Manager ─────────────────────────────────────────────
(node:47875) ExperimentalWarning: CommonJS module /var/home/linuxbrew/.linuxbrew/lib/node_modules/npm/node_modules/debug/src/node.js is loading ES Module /var/home/linuxbrew/.linuxbrew/lib/node_modules/npm/node_modules/supports-color/index.js using require().
Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
npm error code EEXIST
npm error syscall symlink
npm error path ../../../lib/node_modules/npm/man/man5/package-json.5
npm error dest /var/home/linuxbrew/.linuxbrew/share/man/man5/package-json.5
npm error errno -17
npm error EEXIST: file already exists, symlink '../../../lib/node_modules/npm/man/man5/package-json.5' -> '/var/home/linuxbrew/.linuxbrew/share/man/man5/package-json.5'
npm error File exists: /var/home/linuxbrew/.linuxbrew/share/man/man5/package-json.5
npm error Remove the existing file and try again, or run npm
npm error with --force to overwrite files recklessly.
npm error A complete log of this run can be found in: /var/home/marzvrover/.npm/_logs/2024-11-09T05_36_36_758Z-debug-0.log
```

It originally threw the error saying `/home/linuxbrew/.linuxbrew/lib/node_modules/npm/man/man7/package-spec.7` already exists. I renamed that file, but the `package-json.5` file didn't work with the same workaround.

npm log:
```shell
0 verbose cli /var/home/linuxbrew/.linuxbrew/Cellar/node/23.1.0_1/bin/node /home/linuxbrew/.linuxbrew/bin/npm
1 info using npm@10.9.0
2 info using node@v23.1.0
3 silly config load:file:/var/home/linuxbrew/.linuxbrew/lib/node_modules/npm/npmrc
4 silly config load:file:/var/home/marzvrover/.npmrc
5 silly config load:file:/home/linuxbrew/.linuxbrew/etc/npmrc
6 verbose title npm update
7 verbose argv "update" "--location" "global"
8 verbose logfile logs-max:10 dir:/var/home/marzvrover/.npm/_logs/2024-11-09T05_36_36_758Z-
[...]

Issue #1845, opened 2024-10-28
ujust update: poetry update failed with the `rapidfuzz` package


essicolo:
### Describe the bug

When updating with `ujust update`, the poetry update failed with the `rapidfuzz` package.

```
❯ ujust update

── 14:02:10 - System update ────────────────────────────────────────────────────
note: automatic updates (stage) are enabled
Pulling manifest: ostree-image-signed:docker://ghcr.io/ublue-os/bluefin-dx:latest
No upgrade available.

── 14:02:11 - Distrobox ────────────────────────────────────────────────────────

── 14:02:11 - Firmware upgrades ────────────────────────────────────────────────
Metadata is up to date; use --force to refresh again.
Devices with no available firmware updates: 
 • TPM
 • Thunderbolt host controller
Devices with the latest available firmware version:
 • Micron 2200S NVMe 512GB
 • UEFI dbx
Dell Inc. XPS 13 7390
│
└─System Firmware:
  │   Device ID:          c468ffeed9f0dddff00302ba872dee5493082b7f
  │   Résumé:             UEFI System Resource Table device (updated via NVRAM)
  │   Version actuelle:   1.24.0
  │   Version minimum:    1.24.0
  │   Fournisseur:        Dell (DMI:Dell Inc.)
  │   État de mise à jour:Success
  │   GUID:               cbe49cca-492b-93e8-b0f4-f693501f271b
  │   Drapeaux de périphérique:• Périphérique interne
  │                       • Mise à jour possible
  │                       • Le système nécessite une source d'alimentation externe
  │                       • Supported on remote server
  │                       • Needs a reboot after installation
  │                       • Cryptographic hash verification is available
  │                       • Device is usable for the duration of the update
  │   Device Requests:    • Message
  │ 
  └─XPS 13 7390 System Update:
        Nouvelle version: 1.26.0
        Remote ID:        lvfs
        Release ID:       96443
        Résumé:           Firmware for the Dell XPS 13 7390
        Licence:          Propriétaire
        Taille:           16,7 Mo
        Created:          2024-07-05
        Urgence:          Critique
        Fournisseur:      Dell
        Release Flags:    • Trusted metadata
                          • Is upgrade
        Description:      
        This stable release fixes the following issues:
        
        • This release contains security updates as disclosed in the Dell Security Advisories.
        Issues:           DSA-2024-297
                          DSA-2024-243
[...]

PR #1916, opened 2024-11-10
chore(just): restrict topgrade to only supported things

tulilirockz:
This is much more of a RFC than anything concrete, I just wanted to really know what you guys feel like should be directly supported by bluefin (and thus should be added to this `only` array). Restricting topgrade on ujust update should mitigate issues like #1912 and #1845

### DIFF
--- a/system_files/shared/usr/share/ublue-os/topgrade.toml
+++ b/system_files/shared/usr/share/ublue-os/topgrade.toml
@@ -1,5 +1,6 @@
 [misc]
 no_self_update = true
+only = ["system", "flatpak", "distrobox", "firmware"]
 disable = ["self_update", "toolbx", "containers", "helm"]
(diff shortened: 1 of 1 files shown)


tulilirockz:
Brew shouldnt need to be added there since it automatically updates upon running `brew (anything)`. Although it could be a nice idea to automatically update that since maybe users use that very sparringly and they'll have like gigantic updates when theyre trying to install something on brew after a bit of time

m2Giles:
> Brew shouldnt need to be added there since it automatically updates upon running `brew (anything)`. Although it could be a nice idea to automatically update that since maybe users use that very sparringly and they'll have like gigantic updates when theyre trying to install something on brew after a bit of time

brew auto updates with our brew-update services.

castrojo (approved):

PR #1917, opened 2024-11-10
chore(hwe): 09-hwe-additions.sh

m2Giles:
Re-add old surface modules-load file.


castrojo (approved):

macro-stonks:
The changes commited on Nov 10 2024 were after my advice on the Discord that Surface Pro 8 Type cover is non-functioning at LUKS prompt. After a recent update, I am again no longer able to type my LUKS PIN using my type cover. Can this be fixed again?

PR #1923, opened 2024-11-12
fix: CI fixes for ISO/Changelogs

m2Giles:

### DIFF
--- a/.github/workflows/reusable-build-iso.yml
+++ b/.github/workflows/reusable-build-iso.yml
@@ -46,7 +46,7 @@ jobs:
           tar -zxvf just-${JUST_VERSION}-x86_64-unknown-linux-musl.tar.gz -C /tmp just
           sudo mv /tmp/just /usr/local/bin/just
           rm -f just-${JUST_VERSION}-x86_64-unknown-linux-musl.tar.gz
-        
+
       - name: Check Just Syntax
         shell: bash
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/reusable-build-iso.yml
```diff
@@ -59,7 +59,9 @@ jobs:
         shell: bash
         run: |
           image_name="$(just image_name ${{ matrix.base_name }} ${{ matrix.stream_name }} ${{ matrix.image_flavor }})"
-          sudo just build-iso-ghcr "${{ matrix.base_name }} ${{ matrix.stream_name }} ${{ matrix.image_flavor }}"
+          sudo just build-iso-ghcr "${{ matrix.base_name }}" "${{ matrix.stream_name }}" "${{ matrix.image_flavor }}"
+          group=($(groups))
```
m2Giles, line 63: This should be inside the build-iso just recipe

Justfile
```diff
@@ -752,11 +752,9 @@ generate-build-tags image="bluefin" tag="latest" flavor="main" kernel_pin="" ghc
     DEFAULT_TAG=$(just generate-default-tag {{ tag }} {{ ghcr }})
     IMAGE_NAME=$(just image_name {{ image }} {{ tag }} {{ flavor }})
     # Use Build Version from Rechunk
-    BUILD_VERSION={{ version }}
-    if [[ -z "${BUILD_VERSION:-}" ]]; then
-        BUILD_VERSION="${FEDORA_VERSION}.$(date +%Y%m%d)"
+    if [[ -z "${version:-}" ]]; then
+        version="${FEDORA_VERSION}.$(date +%Y%m%d)"
     fi
-    BUILD_VERSION="${BUILD_VERSION:3}"
```
m2Giles, line 759: This means that our dated tags will now be of the style
<stream>-<fedora_version>.<date>

and no longer just <stream>-<date>

This is to work with rechunk's ability to update the version.

m2Giles: this is back in.

castrojo (approved):

PR #1940, opened 2024-11-17
chore(topgrade): make it so topgrade-bootc follows allowlist pattern

tulilirockz:
Should make it so when using LockLayering, the bootc topgrade file will also follow our allowlist from the other topgrade config, seems to work locally
image


bsherman (approved):
This is important to keep the `ujust update` behavior in bluefin the same with or without LockLayering.

Also, I've tested this specific configuration locally and confirm it does what is expected. It only runs the desired upgrade types.

castrojo (approved):

PR #1944, opened 2024-11-18
feat(sddm): add mount unit for installing sddm themes in Discover

RealVishy:
https://github.com/ublue-os/bazzite/pull/1890/files

Copies over this pr from bazzite!


system_files/kinoite/usr/lib/systemd/system/usr-share-sddm-themes.mount
```diff
@@ -0,0 +1,15 @@
...
+[Unit]
+Description=KDE writable sddm workaround
+RequiresMountsFor=/usr/share/sddm/themes
+PartOf=bazzite-kde-themes-workaround.target
+
+[Mount]
+Type=overlay
+What=overlay
+Where=/usr/share/sddm/themes
+Options=lowerdir=/usr/share/sddm/themes,upperdir=/var/sddm_themes/themes,workdir=/var/sddm_themes/themes.work
```
tulilirockz, line 12: This probably mount also should probably automatically create that /var/sddm_themes directory too tho, else it'll be mounting a non-existant folder :p

RealVishy: I've added mkdir command to create the the working directory and the normal directory, I also ended up removing the PartOf= section since we don't want to recreate the bazzite target

build_files/base/07-base-image-changes.sh
m2Giles, line 1: This won't work. You can use a tpmfiles.d conf file to make this directory after boot.

system_files/kinoite/usr/lib/systemd/system/usr-share-sddm-themes.mount
m2Giles, line 1: This also requires mounts for /var

m2Giles (commented):
Take a look at making a symlink to a directory in /etc

Right now this PR won't actually work and has the possibility of failing due to not making sure that /var is mounted as well.

tduck973564:
It would be nice if this could be resolved upstream. Something for me to do, maybe.

build_files/base/07-base-image-changes.sh
RealVishy: Like the one used for the kde-sysmonitor workaround?

m2Giles (approved):

castrojo (approved):

tulilirockz (approved):

PR #1948, opened 2024-11-19
chore(ptyxis): remove profile integration ootb

tulilirockz:
This PR is based off of #1901, should look prettier after it gets merged. (this will also get edited out whenever that happens)

We just need to figure out if we should have a default profile so that we can apply theming to it. Aurora customizes it to have the Breeze look so that it looks a bit more cohesive with the system, but to be completely honestly, none of these profile things worked properly on my system ever, soo......


system_files/shared/usr/share/ublue-os/bluefin-cli/known-ptyxis-profiles
m2Giles, line 1: You can additionally remove this file.

m2Giles (commented):
This looks good for removing everything.

Ptyxis in Aurora sets pallette and opacity via a dconf file. This probably could be moved to a skel file or added to the gschema xml but is inconsequential right now.

castrojo (dismissed):

castrojo (approved):

Issue #1946, opened 2024-11-19
Remove usage of rpm-ostree


tulilirockz:
As you all know, `rpm-ostree` is being "deprecated" in favor of `bootc` and Bluefin should follow that lead, this is just a tracking issue for what I think needs to be done so that we completely do not depend on `rpm-ostree` anymore

- [ ] rpm-ostree cliwrap being used in [main](https://github.com/ublue-os/main/blob/4561ed4858e9a7ad131e7ea2e051ea1a690cff82/install.sh#L18)
- [ ] rpm-ostree being used in [main](https://github.com/ublue-os/main)
- [ ] rpm-ostree calls in the build scripts (e.g.: https://github.com/ublue-os/bluefin/blob/main/build_files/dx/03-packages-dx.sh) (https://github.com/ublue-os/bluefin/pull/1954)
- [ ] rpm-ostree calls in the user-facing ujust & /usr/libexec scripts ([bluefin-system.just](https://github.com/ublue-os/bluefin/blob/main/just/bluefin-system.just) and [bluefin-apps](https://github.com/ublue-os/bluefin/blob/main/just/bluefin-apps.just))

## After removing rpm-ostree/moving to fedora-bootc

- [ ] Add `bootc container lint` to the final build steps

RealVishy:
Would this be as simple as replacing `rpm-ostree` with `dnf` in some of these examples?

RealVishy:
I'm happy to work on the user-facing scripts, once I've got a local registry setup to test on!

castrojo:
This is nowhere close to happening until a bunch of things land upstream, let's investigate down the road.

PR #1954, opened 2024-11-21
chore: un-cliwrap and use dnf5 instead of rpm-ostree

bsherman:
I've been tinkering with and testing dnf5 for a while.

Key items are:
- ensuring cliwrap is properly unwound (checked rpm-ostree Rust code)
- using normal (not wrapped) binaries, eg for dracut
- and the more obvious, replacing "rpm-ostree install" with "dnf5 install", etc
- also uses dnf copr plugin to manage COPRs

Note this does NOT switch to using dnf copr plugin for ublue-os/akmods COPR since doing so removes the RPM managed repo file. That repo file needs to be renamed in the `ublue-os-akmods-addons` package to match the COPR naming convention.

Relates: #1946


tulilirockz:
+1 for F41, it is working perfectly fine indeed

tulilirockz:
Just realized that this will let us use https://github.com/osbuild/bootc-image-builder/ for the ISOs

fiftydinar:
Uncliwrap should be ideally done in main instead, so there would be no need to do it here.

https://github.com/ublue-os/main/pull/679

Also, there would be no need for the script of removing yum, dnf etc.

Only removing `rpm-ostree cliwrap install-to-root /` is enough, as done in main image PR.

Here's screenshot of Silverblue quay base image's output to confirm this
Screenshot From 2024-11-24 13-25-09

bsherman:
I just force pushed an update, and some keen observers may note that I ask people NOT to force-push their PRs... this is true, but in this case there were no comments/requests for changes on any files, so no context was lost.

bsherman:
One known problem with this...

`dnf5` doesn't currently return a failing exit code when a transaction fails. This means it's possible to have failed dnf operations but not fail the build...

This is... undesirable.

I think we need to decide:
1. do we want this now and deal with this risk
-OR-
2. do we wait and see if dnf5 addresses this issue before we get forced to use it as rpm-ostree deprecates functionality in favor of dnf5?

castrojo:
We can just keep this open right? Is there an open upstream issue we can track?

bsherman:
> We can just keep this open right? Is there an open upstream issue we can track?

I have found a few possible issues in the dnf tracker, but I haven't nailed down the exact problem either, so I'm not sure which to track, if any truly relate.

m2Giles:
I think we hold off while we sort out which issues to track.

I've seen some of the issues with dnf5 working with fedora 41 containers and it's very bizarre to get an error message with a 0 exit code.

noelmiller:
> > We can just keep this open right? Is there an open upstream issue we can track?
> 
> I have found a few possible issues in the dnf tracker, but I haven't nailed down the exact problem either, so I'm not sure which to track, if any truly relate.

@bsherman should we open an issue of our own for this in the DNF bug tracker? This would be a pretty serious bug that the DNF team and image mode for RHEL team would want to address. 

Is there any way we can force a DNF command to fail in order to replicate it? Like an intentionally bad mirror or something as a test?

tulilirockz:
https://github.com/ublue-os/bazzite/pull/1905#discussion_r1860797265

bsherman:
> > We can just keep this open right? Is there an open upstream issue we can track?
> 
> I have found a few possible issues in the dnf tracker, but I haven't nailed down the exact problem either, so I'm not sure which to track, if any truly relate.

Possibly upstream issues I've found:
- https://github.com/rpm-software-management/dnf5/issues/1075
- https://github.com/rpm-software-management/dnf5/issues/1862
- https://github.com/rpm-software-management/dnf5/issues/1726

bsherman:
Seems like it is time to bite the bullet on this dnf5 build cutover.

It's been tested a lot in downstream builds. And while there is still some minor concern about a failed install not breaking the build, the benefits seem to outweigh the concerns.

p5 (approved):
LGTM
Been using dnf on my personal image for a few months with zero issues

castrojo (approved):
Awesome, let's merge tomorrow after the weeklies go out. Thanks for this Benjamin, you've slain another multi-year miniboss. Onward to the final boss. 😄

PR #1961, opened 2024-11-23
chore(ci): add 90 day cleanup action

castrojo:
This will clean up images older than 90 days, from bsherman. This has been running fine on kernel-cache.

Only adding a few images, we can add more over time.

### DIFF
--- a/.github/workflows/clean.yml
+++ b/.github/workflows/clean.yml
@@ -1,0 +1,24 @@
+name: Cleanup Old Images
+on:
+  schedule:
+    - cron: "15 0 * * 0" # 0015 UTC on Sundays
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+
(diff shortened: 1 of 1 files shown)


tulilirockz (approved):

tulilirockz:
Would the randomized cron timer interesting for this too? Or is there no reason to use that?

castrojo:
I was thinking running it by hand for a bit until it catches up, then we just leave it on a weekly.

m2Giles:
We should add more than just the base bluefin ones.

We also have:

bluefin-hwe, bluefin-hwe-nvidia, bluefin-surface, bluefin-surface-nvidia, bluefin-asus, bluefin-asus-nvidia and all -dx as well for those and all Aurora versions

castrojo:
Yeah I was planning on adding these, see how it runs before adding in the entire grid.

bsherman (approved):
Can probably do daily but I think this looks fine

PR #1962, opened 2024-11-23
feat(dx): add bootc tooling

tulilirockz:
I would like to add these to `-dx` since they make making images/testing for bootc projects much easier. The osbuild-selinux one is so that the images will get generated w/ the proper SELinux labels, and podman-bootc is so that testing is much easier since it kinda works like `podman machine`. The osbuild-selinux one is particularly useful and non-intrusive

podman-bootc has a brew package but it does not work outside of aarch64 mac :(


tulilirockz:
Note that this add xorriso as a dep https://github.com/containers/podman-bootc/blob/4bab85f45bb85717a9ffd78d4da6515b27e34e23/rpm/podman-bootc.spec#L4

castrojo (approved):
Let's find out what the status is of these landing in distro and open a tracking issue so we don't forget ~F42, thanks!

tulilirockz:
> Let's find out what the status is of these landing in distro and open a tracking issue so we don't forget ~F42, thanks!

By in-distro you mean like, on the base image? - `osbuild-selinux` seems to exist as a package but I hardly think it will be added by default. `podman-bootc` I don't even have any idea what could happen to it. I'll make a tracking issue for these

PR #1965, opened 2024-11-23
chore(flatpaks): move to unified folder

tulilirockz:
This is kind of a nitpick but there is no techinical reason to have that anymore after the justfile rework by m2. This just looks a bit cleaner!


tulilirockz:
Doesn't have any functional changes when building the ISOs and the ujust recipe should work fine also
image

tulilirockz:
Thinking about it - This makes the old images not be able to do `ujust install-system-flatpaks`...

tulilirockz:
If it would be better, I can make it so the `aurora_flatpaks`/`bluefin_flatpaks` folders are still gonna be there until the next GTS build or something

castrojo:
Why are we making this change? Do we really need to change this?

tulilirockz:
Should've proposed this before making the PR LOL - But no, this is most likely not necessary. Should I just close it?

castrojo:
Yeah let's just focus on fixing bugs and clean up, we don't need to rearrange the furniture unless we have to. Thanks!

Issue #1321, opened 2024-05-18
Make stock bluefin less container focused


castrojo:
Note that I mean from a user experience perspective, not a technical level. Now that we have homebrew on the image we can do cooler stuff and gives us some new options that I think we should explore. `bluefin-dx` remains unchanged of course, this will be for our `bluefin` image that we give to our friends. 

First things first I'd like to:

- Remove Containers from the logo menu. 
- Remove boxbuddy
- Strongly de emphasize distrobox and containers in general. 

Let's keep all the container stuff in -dx, there's definitely an audience that loves this brew-first setup, and after trying every combination this one is the one that feels like the sweet spot. I've also totally moved to this model, I keep the devcontainers as part of my IDE and then I have everything else. Feels like we might be overrotating on the need to learn containers, it's supposed to be invisible anyway lol. 

When I need a distrobox we ship 2 keyboard shortcuts to ubuntu and fedora containers, and those get cleaned up every update. And all of those things will still be there, we just won't show them as part of the default experience.

Malix-off:
Are there any GUI to manage homebrew packages?
Also, one-click install `.deb` packages with BoxBuddy felt very user and noob friendly

tulilirockz:
Is this still of interest? Would be a lot easier to do now after the default container purge

tulilirockz:
Everything mentioned here seems to be addressed by now? And boxbuddy also seems like a pretty nice entrypoint for the "how to container" things

castrojo:
I'll close this for now, let's reinvestigate at a later date.

Malix-Labs:
_@castrojo wouldn't closing as not planned be better for now?_

PR #1994, opened 2024-11-30
chore(containers, dx): move remaining container things over to -dx

tulilirockz:
This PR aims to move all the remaining container-related things over to -dx, for now I moved BoxBuddy but idk what else should be moved. (also needs some local testing still.)

Related: #1321

### DIFF
--- a/aurora_flatpaks/flatpaks
+++ b/aurora_flatpaks/flatpaks
@@ -10,7 +10,6 @@ app/org.fkoehler.KTailctl/x86_64/stable
 app/org.kde.haruna/x86_64/stable
 app/org.kde.filelight/x86_64/stable
 app/com.github.tchx84.Flatseal/x86_64/stable
-app/io.github.dvlv.boxbuddyrs/x86_64/stable
 app/io.github.flattool.Warehouse/x86_64/stable
 app/org.fedoraproject.MediaWriter/x86_64/stable
 app/io.missioncenter.MissionCenter/x86_64/stable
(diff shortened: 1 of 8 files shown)


tulilirockz:
M2 on discord: "Leave box buddy"

tulilirockz:
This PR is most likely not needed, just want someone to confirm first tho

m2Giles:
This could be a point of discussion.

Personally, I don't use box buddy; however, it is an extremely approachable way to use distrobox and why I think it should still be on base images.

In -dx l think we have Pods. However, it might make sense to switch over to Podman Desktop since it can manage docker containers too.

tulilirockz:
> This could be a point of discussion.
> 
> Personally, I don't use box buddy; however, it is an extremely approachable way to use distrobox and why I think it should still be on base images.
> 
> In -dx l think we have Pods. However, it might make sense to switch over to Podman Desktop since it can manage docker containers too.

I believe -dx is using podman desktop already tho, Pods got removed at some point afaik

castrojo:
-dx is installing podman-desktop on the ISO already so that part is set. It doesn't get installed when you turn on developer mode, which is a separate problem. Perhaps when/if devpod comes to flathub we should consider installing the flatpaks there, but that's a problem for another day. 

I still really want to do this but it's probably too early and will result in gnashing of teeth.

renner0e:
Would it make sense to also remove Ptyxis from the aurora-non-dx images and replace it with Konsole?
Or at the very least don't hide Konsole and make it the default?

What would be the point in having Ptyxis installed when you don't use the container stuff anyway?
Konsole is way more pleasant to use (on KDE) for anything that is not distrobox.

castrojo:
I'm too tired to argue about ptyxis/kde again, when aurora splits to its own repo they can handle it there.

PR #1997, opened 2024-12-02
fix: remove background-logo extension, restore gnome-tour

castrojo:
This was mentioned in a review and I had forgotten about this thing and we don't use it so removing it.

Also the newer versions of gnome-tour are much nicer than the old ones so restoring this, it's a one time thing for new users.


m2Giles:
My funny activate Linux will be gone. But done

m2Giles (approved):

PR #2008, opened 2024-12-05
fix(accent-color): light theme for accent colored motd

tulilirockz:
This fixes a regression with the previous accent color PR that made it so it always thought it was a dark theme. This should work fine now!


tulilirockz:
image

castrojo (approved):

PR #2014, opened 2024-12-06
feat: allow non-official image repositories in ublue-rollback-helper

cr7pt0gr4ph7:
Use the image registry that was used when building the currently running image, instead of unconditionally switching to `ghcr.io/ublue-os`.

See also #2016.


castrojo (approved):

cr7pt0gr4ph7:
Seems like a transient build failure due unrelated updates in a package repository, and not related to this PR. Can you attempt to merge again?

PR #2019, opened 2024-12-08
fix(motd): use bold for color blocks instead of background color

tulilirockz:
This is kind of a workaround but it works perfectly well. It just makes it so light and dark themes look the same!!! LES GOOOO this should be the last fix for the MOTD


tulilirockz:
image
image

castrojo (approved):

PR #2029, opened 2024-12-08
chore(feat): adjust cron timers

castrojo:
This adjusts bluefin's crons another +5m. This means bazzite, aurora, and bluefin will pack the queue in relative serial order.


m2Giles:
This needs to make sure that the workflow calls have matching times for the schedule conditions checks.

castrojo:
Do you mean matching them all? So like 5:50 for all of them?

m2Giles (approved):
Looks like that's all of them.

PR #2033, opened 2024-12-09
fix: remove aurora from changelogs

castrojo:
Please advise on line 117


tulilirockz:
I think you can just remove the aurora reference from there? Just hard code the image to return "bluefin" instead

m2Giles:
This needs some more cleanup.

GTS and normal Image Matrix are now the same so should be merged.

Aurora image is still part of the generator here: https://github.com/ublue-os/bluefin/blob/3cbe746751f683e77e2029a9e65c9d50b09b5f0c/.github/changelogs.py#L117

m2Giles (approved):

PR #2052, opened 2024-12-16
chore: workaround podman image scp leaving behind temporary files in /var/tmp

detiber:
In at least some cases, it seems that podman leaves behind temporary files created during the `podman image scp` process.

Since this is used by both the rechunk and build-iso just recipes when using locally built images, this can result in a lot of disk space used when testing locally.

Creating a temporary directory, overriding TMPDIR when calling `podman image scp`, and then removing that temporary directory works around the issue.


bsherman (approved):
Looks good to me.

Should be harmless at worst, and the value is clear at best.

@m2Giles correct me if I'm wrong.

m2Giles:
Looks fine.

PR #2062, opened 2024-12-19
feat(ci): use just action for installation

castrojo:
Tip from detiber, as recommended by the just docs.


m2Giles:
Doesn't work since not on Root's PATH.

tulilirockz (approved):

tulilirockz:
oh damn

PR #2067, opened 2024-12-21
refactor: move image signature verification to Podman

p5:


p5:
This doesn't work as I expected it to.

PR #2076, opened 2024-12-25
chore: move wallpapers to COPR build instead of having it on repo

tulilirockz:
Requires https://github.com/ublue-os/packages/pull/68


tulilirockz:
Just if someone is checking this out: We are testing this on [achillobator](
image
) currently. This should work just fine from my findings LOL

castrojo (approved):

PR #2090, opened 2025-01-01
chore(build-dx): Clean up pretty name in OS release info

EyeCantCU:
Remove -dx from the pretty name in OS release info to consolidate brew stats given it is a mode and not a variant

### DIFF
--- a/build_files/shared/build-dx.sh
+++ b/build_files/shared/build-dx.sh
@@ -28,7 +28,7 @@ sysctl -p
 /ctx/build_files/dx/04-override-install-dx.sh
 
 # Branding Changes
-sed -i '/^PRETTY_NAME/s/Bluefin/Bluefin-dx/' /usr/lib/os-release
+sed -i '/^PRETTY_NAME/s/Bluefin/Bluefin DX/' /usr/lib/os-release
 
 # Systemd and Disable Repos
 /ctx/build_files/dx/09-cleanup-dx.sh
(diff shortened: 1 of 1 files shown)


build_files/shared/build-dx.sh
```diff
@@ -28,7 +28,7 @@ sysctl -p
 /ctx/build_files/dx/04-override-install-dx.sh
 
 # Branding Changes
-sed -i '/^PRETTY_NAME/s/Bluefin/Bluefin-dx/' /usr/lib/os-release
+sed -i '/^PRETTY_NAME/s/Bluefin/Bluefin DX/' /usr/lib/os-release
```
castrojo, line 31:
```suggestion
sed -i '/^PRETTY_NAME/s/Bluefin/Bluefin/' /usr/lib/os-release
```

castrojo (changes requested):
How about this so our brew stats get consolidated? -dx shouldn't be a sku it's more of a mode.

castrojo (dismissed):

build_files/shared/build-dx.sh
EyeCantCU: Makes perfect sense to me, let's remove the line entirely since it isn't doing anything anymore :)

castrojo: oh, derp!

castrojo (approved):

Issue #2082, opened 2024-12-29
Bluefin GTS installer missing udev module


cablespaghetti:
### Describe the bug

Related to https://github.com/JasonN3/build-container-installer/issues/151

I downloaded the latest GTS image today and tried to boot the installer on an old laptop (BIOS not UEFI if it’s significant).

Anaconda failed to start due to a missing udev Python module.

Here is a “screenshot”
![Image](https://github.com/user-attachments/assets/c745b8e5-75c6-42ac-a21d-31d81b6fd8aa)

### What did you expect to happen?

The anaconda installer starts so I can install my Bluefin goodness.

### Output of `bootc status`

```shell

```

### Output of `groups`

```shell

```

ledif:
Looks like there's already a PR in the works to fix this in the upstream `build-container-installer` action: https://github.com/JasonN3/build-container-installer/pull/152

castrojo:
Investigating now, what image did you download? I can get you a backup link to an older ISO in the meantime.

cablespaghetti:
No stress I'm happily running stable on the laptop now. Just the basic GTS image.

castrojo:
I'm copying over the known-good backups into the cloudflare bucket now while we investigate the installer, thanks for catching this!

This should take a few minutes, I'll update this issue when the copy is complete, then a redownload should do the trick, thanks!

castrojo:
Ok copy complete, if you wouldn't mind giving it a shot now!

cablespaghetti:
I think you've copied over a Fedora 39 based image. Is that by design?

castrojo:
It's the backup we have, after installation it'll upgrade you to the right version (either automatically or you can run the update manually). We've confirmed this is a GTS-only problem. 

I've added a note on the website in the meantime. We'll freshen them up when we find a fix!

cablespaghetti:
Sounds good. Maybe I can be useful and make a PR to store some number of historical images in the build pipeline depending on your appetite for cloud storage bills?

castrojo:
Yeah it's about 4 bucks now, wouldn't hurt to hold another set or two! We'd only need this in `gts` and `stable`. 

We make ISOs for `latest` but don't post them on the website so those we can probably remove them entirely. 

Also might be prudent to just do a run of them during a release instead of building them weekly.


cc @m2Giles

castrojo:
All new ISO process will take care of this, working on it for F42

PR #2093, opened 2025-01-02
chore(ci): Temporarily disable ISOs

m2Giles:


castrojo:
Note for the record: this is to investigate: https://github.com/ublue-os/bluefin/issues/2082

PR #2096, opened 2025-01-02
docs: update motd

castrojo:
bluefin-cli is opt-in - The way it reads now it implies that everyone is using it, this will make it clearer I think.


tulilirockz (approved):
haha yeah it makes more sense. would be best to go with that

p5 (approved):

m2Giles (approved):

PR #2106, opened 2025-01-04
 fix(actions): try to recover if just fails to install

ledif:
I have noticed an increase in workflow failures due to flaky errors with installing `just`. For example, in the past few days, the following jobs to build images have failed:

- https://github.com/ublue-os/bluefin/actions/runs/12600726942/job/35120374239
- https://github.com/ublue-os/bluefin/actions/runs/12576191989/job/35052038346
- https://github.com/ublue-os/bluefin/actions/runs/12567338911/job/35033772133

The command `curl -L https://api.github.com/repos/casey/just/releases/latest | jq -r '.tag_name'` is sometimes returning "null" instead of an actual version, but it's difficult to tell if the curl call is failing or if it's returning a payload that is being parsed incorrectly.

In either case, adding `--retry` to curl and if it's still null, hardcoding a known good version should do the trick for now.

### DIFF
--- a/.github/workflows/generate-release.yml
+++ b/.github/workflows/generate-release.yml
@@ -38,14 +38,8 @@ jobs:
 
       - name: Install Just
         shell: bash
-        run: |
-          set -eoux pipefail
-          JUST_VERSION=$(curl -L https://api.github.com/repos/casey/just/releases/latest | jq -r '.tag_name')
-          curl -sSLO https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-x86_64-unknown-linux-musl.tar.gz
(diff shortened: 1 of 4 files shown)


tulilirockz:
I believe we should try using that [setup-just](https://github.com/centos-workstation/achillobator/blob/5cf1fa16762b29676e4752ca0bf181f9f6344d7a/.github/workflows/build-image.yml#L42) action at some point, but this is a very nice idea

m2Giles:
This should be more resistant

```bash
while [[ "${JUST_VERSION:-}" =~ null || -z "${JUST_VERSION:-}" ]]
do
    JUST_VERSION=$(curl -L https://api.github.com/repos/casey/just/releases/latest | jq -r '.tag_name')
done
```

m2Giles (approved):

PR #2112, opened 2025-01-05
feat: add artifacthub metadata on rechunker step

tulilirockz:


m2Giles:
I'm fixing the justfile to not pass things to `just sudoif` so the labels can be more easily applied.

I think I got most of these in my PR but probably don't have some of them right.

tulilirockz:
@m2Giles mind pinging me whenever you do that? would be super nice to have that on aurorazzitebluefin

PR #2115, opened 2025-01-06
fix: move powertop to all images

castrojo:
People without dx should be able to use this.


ledif:
Does it make sense to remove `powertop` from the `dx` group as well, since now it's in there twice?

tulilirockz (approved):

PR #2119, opened 2025-01-07
fix(pin): sssd to 2.10

### LINKED ISSUE #2118: Pin SSSD version at 2.10.0 for stable
### Describe the Package
Security package that facilitates kerberos authentication for domain logins for active directory accounts.

SSSD packages for version 2.10.1 block the ability for enterprise logins on Fedora 41. If these could be pinned on stable that would be great, Latest could continue updated so that we can check when the newer updated packages begin working again.

### Bodhi Link (Optional)

Pin to this version please: https://bodhi.fedoraproject.org/updates/FEDORA-2024-bfd5344277

castrojo:
Fixes #2118 

See also: https://universal-blue.discourse.group/t/bluefin-enterprise-woes/6157/4

### DIFF
--- a/build_files/base/18-workarounds.sh
+++ b/build_files/base/18-workarounds.sh
@@ -19,4 +19,11 @@ fi
 #    rpm-ostree override replace https://bodhi.fedoraproject.org/updates/FEDORA-2024-dd2e9fb225
 #fi
 
+# Ping sssd version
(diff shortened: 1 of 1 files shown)


p5 (approved):

bsherman (approved):
Probably need a long term goal here... but maybe the long term play is let Fedora figure out things before unpinning.

PR #2128, opened 2025-01-09
feat: move fastfetch config (and logos) to COPR package

tulilirockz:
Needs some testing, but this should be it


tulilirockz:
That `bluefin-fastfetch` package is just for the fastfetch configuration, `ublue-fastfetch` that is the crux of it

castrojo (dismissed):

p5 (approved):

Issue #2046, opened 2024-12-14
Readd devcontainer support


GarciaLnk:
The previous devcontainer went missing since the build process was reorganized, I suggest readding a devcontainer that eases off the development process when it's done on a non-bluefin host.

Since podman is not very well-supported on devcontainers it may be needed to modify the Justfile workflow to allow the usage of docker in the build process.

m2Giles:
That's why it was removed. The justfile was designed to use tools that will be on every ublue image. It uses bash, coreutils, curl, skopeo, jq, and podman (python for changelogs only). Devcontainers don't work well with podman. Additionally for rechunk and the ISO build it requires a rootful container with quite a bit of privileges. While it would work if using a privileged docker container, I'm unsure if shipping that as a recommended solution makes the most sense.

It also would require quite a bit of rework of the justfile to handle using docker instead and I would be concerned about code rot given what happened with the previous version of the justfile.

That said, I'm not 100% against it, but it will require a bit of work.

GarciaLnk:
Given that the devcontainer would only be used for local development, we could omit the rechunk step for that use case, being able to build an image and an .iso inside a devcontainer would be enough imo.

m2Giles:
iso and rechunk both need privileges.

m2Giles:
It is now possible to use the just actions with docker now using changes from #2114. If podman doesn't exist it will attempt to use docker or docker can be explicitly specified with an environment variable.

This means that a docker outside of docker devcontainer could be used. For docker inside of docker, you would need to use a privileged container (which I believe is the default for docker in Docker).

PR #2131, opened 2025-01-09
feat: add devcontainer

GarciaLnk:
Fixes https://github.com/ublue-os/bluefin/issues/2046

I've confirmed that the following recipes are working properly inside the devcontainer:

```
just build
just run
just build-iso
just run-iso
```

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,37 @@
+// For format details, see https://aka.ms/devcontainer.json.
+{
+	"name": "Bluefin-devcontainer",
+	"image": "mcr.microsoft.com/devcontainers/base:ubuntu",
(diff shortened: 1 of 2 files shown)


m2Giles:
Also need to check rechunk. But it should work with docker out of docker.

Did you try Docker in Docker? I think this one is probably a footgun with people trying to do this inside of a rootless podman container though.

GarciaLnk:
> Also need to check rechunk. But it should work with docker out of docker.

Rechunk fails with docker because `docker mount` doesn't exist.

> Did you try Docker in Docker? I think this one is probably a footgun with people trying to do this inside of a rootless podman container though.

Docker in docker throws a bunch of permissions errors, I didn't dig too deep though.

castrojo (approved):
Thank You!

PR #2135, opened 2025-01-11
chore: move homebrew configuration to ublue-brew

tulilirockz:


tulilirockz:
Might need to remove `brew.sh` from this package since ublue-os-just breaks it right now

```
error: Error running transaction: file /etc/profile.d/brew.sh from install of ublue-brew-0.1.0-1.fc41.x86_64 conflicts with file from package ublue-os-just-0.37-1.fc39.noarch
  ```

castrojo (dismissed):

p5 (approved):

PR #2136, opened 2025-01-11
chore: switch from ublue-update to uupd as default automatic updater

tulilirockz:
This should remove all references to ublue-updater and it's files from the system

We need to be very cautious about this PR since it might break automatic updates for people. And that would suck very much.


castrojo:
Yeah let's hold this one for a while. Let's let achillobator do the work on this.

castrojo (approved):

bsherman (approved):

bsherman (approved):

system_files/shared/usr/share/ublue-os/just/10-update.just
gerblesh, line 1: Not sure if we should use `uupd` for manual updates with ujust, I opted not to because it's not very interactive and we can't display bootc progress yet.

m2Giles:
I'm considering turning this on for beta.

I also think we can maybe have a service that checks the status of uupd. If uupd fails, it runs `rpm-ostree upgrade` and hopefully catches us if we have some edge cases.

Again I haven't had issues with uupd.

I also agree that we should not change ujust update at this time. Again to help give us padding.

castrojo:
+1 for enabling on beta asap!

m2Giles:
@tulilirockz what do you think of my changes?

It is only going to install on beta for now and the enable/disable will be transparent to the user.

tulilirockz:
Sounds good to me! I also think that just having `uupd` as a background thing is best.

ledif (approved):
LGTM

gerblesh (approved):
LGTM

castrojo (approved):

PR #2150, opened 2025-01-21
feat: Update VFIO to be bootc compliant

HikariKnight:
This will be kept as a draft in-case more changes are needed, but from what i can tell, dracut is being run as part of the build process so the vfio modules should be included into initramfs with the image with this inclusion.

the just recipes and the script for kvmfr has been updated to reflect this

also minor changes to heredocs in some recipes to make the just highlighting work for me in vscode when writing this.
I will be making a similar PR to Aurora with the same updates later.

### DIFF
--- a/just/bluefin-system.just
+++ b/just/bluefin-system.just
@@ -180,8 +180,8 @@ configure-vfio ACTION="":
                 sudo chmod +x /etc/libvirt/hooks/qemu
                 sudo grep -A1 -B1 "# Add" /etc/libvirt/hooks/qemu | sed 's/^# //g'
                 echo ""
-                if sudo test ! -d "/etc/libvirt/hooks/qemu/qemu.d"; then
-                    sudo mkdir /etc/libvirt/hooks/qemu/qemu.d
+                if sudo test ! -d "/etc/libvirt/hooks/qemu.d"; then
+                    sudo mkdir /etc/libvirt/hooks/qemu.d
                 fi
             fi
             echo "VFIO will be enabled on next boot, make sure you enable IOMMU, VT-d or AMD-v in your BIOS!"
--- a/just/bluefin-system.just
+++ b/just/bluefin-system.just
@@ -35,11 +35,11 @@ toggle-devmode:
     source /usr/lib/ujust/ujust.sh
     CURRENT_IMAGE=$(rpm-ostree status -b --json | jq -r '.deployments[0]."container-image-reference"')
     if grep -q "/var/ublue-os/image" <<< $CURRENT_IMAGE ; then
-        cat <<EOF
+        bash -c "cat <<EOF
     Before we can switch to the Bluefin Developer Experience
     the current system needs an update. Please run 'ujust update'
     and reboot your system when the update is finished
-    EOF
+    EOF"
         exit
     fi
     if /bin/grep -q "\-dx" <<< $CURRENT_IMAGE ; then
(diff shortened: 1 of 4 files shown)


just/bluefin-system.just
```diff
@@ -195,9 +195,6 @@ configure-vfio ACTION="":
               echo "$VFIO_IDS"
               VFIO_IDS_KARG="--delete-if-present=\"$VFIO_IDS\""
             fi
-            echo "Removing dracut modules"
-            sudo rm /etc/dracut.conf.d/vfio.conf
-            rpm-ostree initramfs --enable
             rpm-ostree kargs \
```
tulilirockz, line 198: Can we still run kargs on a bootc-only system?

system_files/dx/usr/libexec/bluefin-dx-kvmfr-setup
```diff
@@ -15,17 +15,22 @@ if [ "$CONFIRM" == "Cancel" ]; then
...
+# Default value set by us is 128 which is enough for 4k SDR
+# Find the current value by running "rpm-ostree kargs"
+KVMFR_MODPROBE'
 
-# Add kvmfr modprobe file following upstream documentation
-sudo bash -c "cat << KVMFR_MODPROBE > /etc/modprobe.d/kvmfr.conf
-options kvmfr static_size_mb=128
-KVMFR_MODPROBE"
+# Add kvmfr static size karg
+rpm-ostree kargs --append-if-missing="kvmfr.static_size_mb=128"
```
tulilirockz, line 33: Using the modprobe override here seems like what we should do with bootc, but maybe it just works?

just/bluefin-system.just
HikariKnight: we use it in bazzite, bootc just does not like the initramfs rebuilding and layering from my understanding.
but i can remove lact from my install and double check to be 100% sure

HikariKnight: just checked, can be used with kargs, just not layers and initramfs

system_files/dx/usr/libexec/bluefin-dx-kvmfr-setup
HikariKnight: we probably can and the looking-glass people would most likely prefer it, i am just following the change kyle did in bazzite to add the vfio modules to the image. :)

HikariKnight:
putting back to draft as i remembered there is a part i forgot to move over from bazzite, so will work on that later today or in the weekend hopefully

system_files/dx/usr/libexec/bluefin-dx-kvmfr-setup
HikariKnight: thinking back at this, unsure if this will require rebuilding the initramfs when the user changes it though which is what we are trying to avoid.

tulilirockz:
Is this good for merging? It looks good to me but I cant really test this on my machine.

system_files/dx/usr/libexec/bluefin-dx-kvmfr-setup
HikariKnight: Just checked, we went with a karg as modprobe is ignored due to initramfs not being regenerated.

HikariKnight:
Should be good to merge, but i will run a last test pass over it in a VM after i get home from work today.

i will manually delete the branch after it is merged as i have to copy things over to aurora.

HikariKnight:
@tulilirockz all cleaned up now :)

tulilirockz (dismissed):
Nothing seems wrong, if this causes issues people are gonna report it so its ok if its not working perfectly yet.

HikariKnight:
Ported changes to aurora too, so safe to delete branch once merged

tulilirockz (approved):
To be honest I dont really have much of a way to test this one as I dont personally use VFIO, but it looks good to me, code seems fine and the `ls-iommu` stuff looks good too

PR #2151, opened 2025-01-23
fix(artifacthub): remove artifacthub category

castrojo:
This category doesn't exist, we need to remove it or it fails to ingest in ah.


castrojo:
I don't understand how I broke the just file by removing these 2 lines?

p5 (approved):

ledif:
The `just check` has been wonky lately. It doesn't help that there's a mismatch of versions between what is shipped in the images and what we use to validate our justfiles in the GH actions.

castrojo:
I'm assuming there's version skew between just in fedora and just-the-action?

ledif:
Try pulling in the changes to Justfile here: https://github.com/ublue-os/aurora/pull/144/files.

m2Giles:
You didn't. The formatter is slightly different on the newest version of just (which is used during the action) compared to what we have from fedora repos.

PR #2155, opened 2025-01-24
feat: remove inputleap from base image

dreamyukii:
InputLeap is now available as flatpak package on [flathub](https://flathub.org/apps/io.github.input_leap.input-leap).


castrojo:
Is there a migration/announcement plan?

dreamyukii:
i think we should announce this on Discourse and Discord first for Bluefin/Aurora

PR #2163, opened 2025-01-28
fix: allow wheel users to access yubikeys

rothgar:
This is from the Yubikey guide
https://github.com/drduh/YubiKey-Guide/issues/376

It allows users in the wheel group to access gpg signing card


tulilirockz:
Sweet! Might be nice adding this to the shared packages (maybe). Considering this might diverge between Centos and Fedora it might not be a good idea

castrojo (approved):
This should be in config but landing stuff there is impossible lately so snagging it.

rothgar:
I forgot to do this last year and finally reinstalled and it bit me again. Decided to put it somewhere this time.

tulilirockz (approved):

PR #2165, opened 2025-01-28
fix: disable multimedia repos before Nvidia install

p5:


dylanmtaylor:
This should be unnecessary as the two are now in sync.

inffy:
> This should be unnecessary as the two are now in sync.

This will happen pretty much everytime when nvidia updates drivers. Like now, drivers were updated yesterday and negativo repos are out of sync (multimedia has newer driver than the nvidia repo)

bsherman:
I think we can close this one @p5 as we're addressing this issue in other ways.

Issue #2157, opened 2025-01-26
stable-20250126.1 is unbootable on Nvidia systems


Malix-Labs:
### Describe the bug

stable-20250126.1 seems unbootable (tested with Bluefin-DX-Nvidia on a Lenovo Legion 5)

My booting screen (personally, with a Lenovo and a Bluefin logo plus a spinning loading icon) shows up, then seems to freeze after a few seconds, then the screen turns full black forever

I have to power off from long pressing the power button, and select the previous deployment (41.20250119.1)

### What did you expect to happen?

A working stable release

### Output of `rpm-ostree status`

```log
❯ rpm-ostree status
State: idle
AutomaticUpdates: stage; rpm-ostreed-automatic.timer: no runs since boot
Deployments:
  ostree-image-signed:docker://ghcr.io/ublue-os/bluefin-dx-nvidia:stable
                   Digest: sha256:3a2bd576580c679d83174e8c2391108d26920cdeea44ffd8bede85ebedd071db
                  Version: 41.20250126.1 (2025-01-26T06:07:45Z)
            SecAdvisories: 1 low, 2 important, 1 critical
                     Diff: 162 upgraded, 1 removed, 1 added
          LayeredPackages: google-chrome-beta waydroid

● ostree-image-signed:docker://ghcr.io/ublue-os/bluefin-dx-nvidia:stable
                   Digest: sha256:dfdbfb5acd4390ed61d024aaa356568db90c20aff5c75ae76724a344f2f95cae
                  Version: 41.20250119.1 (2025-01-19T06:13:33Z)
          LayeredPackages: google-chrome-beta waydroid
```

### Output of `bootc status`

```log
❯ bootc status
ERROR Status: Initializing storage: Acquiring sysroot: Preparing for write: Querying root privilege: This command requires root privileges
```

### Output of `sudo bootc status`

```log
❯ sudo bootc status
No staged image present
Current booted state is native ostree
Current rollback state is native ostree
```

### Output of `fastfetch`

```log
❯ fastfetch
                                               ▖      malix@malix-pc
                                   ▃▃▄▃▂▃▄▅▅▅▂▄▆
                                ▗▄▄▅╱▅▅▆┯▘┳▇▗▄┱╴    󱋩  bluefin-dx-nvidia:stable 🔐
[...]

PR #2168, opened 2025-01-29
feat: update bluefin stable kernel pin to 6.12.9

bsherman:
Note: 6.12.9 is the current coreos-testing kernel and is known to have fixed the AMD suspend bug.

We're switching the "bluefin:stable" builds to use this kernel since it allows us to ship not-broken kernel and not-broken nvidia.

We don't have a pre-existing build for Fedora 40, so we can't ship a fix for "bluefin:gts" at this time.

Partial fix for #2157

This fixes `bluefin:stable`  but  not `bluefin:gts`

### DIFF
--- a/.github/workflows/build-image-gts.yml
+++ b/.github/workflows/build-image-gts.yml
@@ -6,8 +6,8 @@ on:
       - testing
     paths-ignore:
       - "**.md"
-  schedule:
-    - cron: "50 5 * * 0" # 5:50 UTC Weekly on Sundays
+  # schedule:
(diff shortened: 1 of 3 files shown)


bsherman:
@castrojo maybe this PR should include a disabling of the GTS build until we have the officicial updated kernel?

bsherman:
Looks like this is building what we expect for stable... a 6.12.9 kernel with nvidia 570 driver.

p5 (approved):
Thank you!

PR #2169, opened 2025-01-29
chore: use solaar flatpak instead of the system package

tulilirockz:
It works out of the box now! Should probably follow a similar path to https://github.com/ublue-os/bluefin/pull/2155


castrojo:
Yeah I think both of these are really user visible, so we should just do these when F42 hits so we can put it in the release announcement. Would prefer to not interrupt the userbase during the cycle.

tulilirockz:
Yup, probably best way to get that going

PR #2199, opened 2025-02-10
fix: only run ublue user services for human users

ledif:
Porting changes from https://github.com/ublue-os/aurora/pull/206 to here.

I'm not sure if GDM has the same design as SDDM, which would be an issue for GDM. In any case, we are still running these user services for the root user.

### DIFF
--- a/system_files/dx/usr/lib/systemd/user/bluefin-dx-user-vscode.service
+++ b/system_files/dx/usr/lib/systemd/user/bluefin-dx-user-vscode.service
@@ -2,6 +2,7 @@
 Description=Run 
 Wants=network-online.target
 After=network-online.target ublue-user-setup.service
+ConditionUser=!@system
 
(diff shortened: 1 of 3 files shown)


ledif:
@tulilirockz Should I delete the `system_files/shared/usr/lib/systemd/user/ublue-user-setup.service` file? I noticed that this file already exists in the packages repo in https://github.com/ublue-os/packages/tree/main/packages/ublue-setup-services/src/user-services

tulilirockz:
@ledif It does! Its just that that package is consumed differently than these scripts, having it here would be a major rework for another time :)

castrojo:
(bot testing)

@dosu can you summarize this pull request for me?

tulilirockz (approved):
This looks good to me. We should eventually just move to the `ublue-setup-services` package directly anyways

Issue #2190, opened 2025-02-07
Cleanup fonts


GarciaLnk:
Currently, we are shipping 4 typefaces using their font files (https://github.com/ublue-os/bluefin/tree/main/system_files/shared/usr/share/fonts), out of these, we already ship Ubuntu fonts through an RPM in the DX image (ubuntu-family-fonts) and Inter is available as a Fedora package (rsms-inter-fonts). We should avoid shipping .ttf files since they will become outdated.

I'm unsure how relevant Meslo and Commit are, but we can't ship every typeface under the sun else the image would be a few GB larger (and we already ship 11 different fonts in the DX images, excluding the nerd-fonts fallback).

castrojo:
Yeah what we need to do is figure out exactly what to ship. Want to take lead on this? We don't need to ship everything but we should have the popular ones. We can probably save some space and drop some external dependencies too.

PR #2200, opened 2025-02-10
chore: cleanup fonts

GarciaLnk:
This PR removes from the image all the fonts that can be installed with Embellish, while keeping those used in default configs (such as JetBrains Mono or Cascadia Code). It also installs Inter as an RPM.

I've been as aggressive as I can be removing fonts without affecting the default experience, I'm open to modifying the PR if one font is particularly popular.

Closes https://github.com/ublue-os/bluefin/issues/2190


castrojo:
This just removes all the fonts? I thought we were moving the fonts to RPMs instead of individual files?

GarciaLnk:
I didn't see the point in having half a dozen additional fonts in DX when we are already shipping Embellish, but as I said, I'm open to restoring in this PR whatever fonts are most popular.

castrojo:
Yes we still want to ship the fonts, "A collection of well-curated monospace fonts" is one of the major features of DX, and the ubuntu font in particular is part of our identity.

PR #2204, opened 2025-02-12
feat: Fix geolocation (Primarily for timezone detection) now that Mozilla has exited

### LINKED ISSUE #2203: Configure geoclue to use beacondb
We can't set timezones automatically due to mozilla's geoclue service not existing. There's an alternative: 

https://fedoramagazine.org/the-state-of-the-location-permission-on-fedora-linux-in-2025/

IMO we should set this up for bluefin and bazzite's gnome images so geoclue works.

KyleGospo:
Fixes #2203

### DIFF
--- a/system_files/shared/etc/geoclue/conf.d/99-beacondb.conf
+++ b/system_files/shared/etc/geoclue/conf.d/99-beacondb.conf
@@ -1,0 +1,3 @@
+[wifi]
+enable=true
+url=https://api.beacondb.net/v1/geolocate


castrojo:
+1, let's see how it works in bazzite's testing branch. Thanks!

system_files/shared/etc/geoclue/conf.d/99-beacondb.conf
```diff
@@ -0,0 +1,3 @@
+[wifi]
+enable=true
+url=https://api.beacondb.net/v1/geolocate
```
jplock, line 3: Is this a valid URL?

```
curl https://api.beacondb.net/v1/geolocate -i
HTTP/2 404 
server: nginx
date: Wed, 12 Feb 2025 23:50:58 GMT
content-length: 0
```

KyleGospo: It should be, this entire thing is from the Fedora magazine.

p5: It works using a POST request.  GETs return a 404

PR #2223, opened 2025-02-25
fix: ujust changelog shows stable channel even when on GTS

### LINKED ISSUE #2221: ujust changelog shows stable channel even when on GTS
### Describe the bug

On Bluefin GTS the `ujust changelog` command shows me information on recent changes to the stable channel, which is irrelevant when on the GTS channel.

Reported on the forum already, but copying to a GitHub issue as requested: https://universal-blue.discourse.group/t/ujust-changelog-showing-updates-for-stable-instead-of-gts/7052

I expect the changelog command to by default show me changes for my current channel.

### Output of `bootc status`

```shell

```

### Output of `groups`

```shell

```

_No response_

mheuvel-dev:
Fixes #2221 by picking the correct channel and by selecting the currently deployed version tag.

### DIFF
--- a/just/bluefin-apps.just
+++ b/just/bluefin-apps.just
@@ -132,7 +132,6 @@ install-jetbrains-toolbox:
 incus:
     @ujust install-incus
 
-
 # Install OpenTabletDriver, an open source, cross-platform, user-mode tablet driver
 [group('Apps')]
 install-opentabletdriver:
@@ -164,7 +163,6 @@ install-opentabletdriver:
     else
         echo "Have a good day :)!"
     fi
-
 
 # Install and configure Incus
 [group('Apps')]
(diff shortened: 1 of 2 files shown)


tulilirockz:
Tested this out on a `:latest` build, doesnt seem to be working...
image

mheuvel-dev:
That's correct. I should have commented on that.

There are no release tags for latest and daily images. The previous "latest" was pointing to the latest stable. Hence, if you are on a latest channel, then it was showing you a irrelevant changelog. Now it is failing. A solution is to tag those images.

mheuvel-dev:
Maybe we should make the error prettier?

m2Giles:
Honestly think the easiest is to fallback to the previous changelogs.

We can't readily bake the changelogs into the image and the script takes a bit of time to run. So just fallback to the other changelogs.

castrojo (approved):

PR #2226, opened 2025-02-27
chore: pin latest to last 6.12 kernel

bsherman:
This pins bluefin latest kernels to our newest good builds:
- main/nvidia flavors: 6.12.15
- hwe/hwe-nvidia flavors: 6.12.12.bazzite

These are the newest known good kernels for the respective flavors.

The purpose of the pin is avoid shipping kernel 6.13 until the major issues are resolved (eg "bad page state" and fuse issues crashing systems and flatpaks).

### DIFF
--- a/.github/workflows/build-image-latest-hwe.yml
+++ b/.github/workflows/build-image-latest-hwe.yml
@@ -1,4 +1,4 @@
-name: Latest Images
+name: Latest hwe Images
 on:
   merge_group:
   pull_request:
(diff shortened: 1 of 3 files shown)


m2Giles:
This is a good improvement.

bsherman:
It looks like this builds properly.

It won't hurt us to pin latest and wait to make sure the new kernels are good.  Fedora is working on a 6.13.5 which is supposed to fix the huge concerns users are experiencing, but probably better to be safe. Plus, our HWE users are already on bad 6.13 kernels, so we really should help them out.

castrojo (approved):

PR #2228, opened 2025-03-01
feat: add justfile recipe to retag broken NVIDIA ghcr images

ledif:
This is an adaptation of the same recipe used in Aurora that we use to retag broken NVIDIA images.


m2Giles:
This only does bluefin/bluefin-dx and does not do HWE images.

m2Giles (approved):
See, note. This only does bluefin/bluefin-dx main images. It does not work for hwe.

This could  be handled via another just recipe.

Issue #2242, opened 2025-03-10
March wallpaper doesn't render properly in GTS


castrojo:
The wallpaper in GTS is broken, shows up as a grey box, we should check the jxl file. This is minor but very visible so we should check the files. 

Works fine in LTS and stable.

castrojo:
(bot testing)

@dosu how can I verify the integrity of a jxl file?

PR #2243, opened 2025-03-10
fix: remove march wallpaper from GTS (f40)

tulilirockz:
This should pull in an old version of the bluefin wallpapers package, so we dont get the black screen issue since... we dont get the new wallpaper at all!

Fixes: #2242


p5 (commented):
I think we also need to change the build scripts to not inject 03-bluefin into the gsettings/dconf

bsherman (approved):

PR #2278, opened 2025-03-21
chore: get beta building....

m2Giles:


p5 (approved):

m2Giles:
This has added quite a bit of guards/skips to get beta building.

So far we have:
1. Ripped out docker
2. Put guards on all akmods installs. I believe only necessary for openrazer right now.
3. Add --skip-unavailable flag for packages.sh installs from our json
4. Skips swaps for patches to fwupd, gnome-shell, and switcheroo-control.

castrojo (approved):

PR #2281, opened 2025-03-23
chore: simplify packages.sh install scripts

m2Giles:
This cleans up the packages.json removing the silverblue objects. Also fixed alphabetical ordering.

This also stops -dx from attempting to reinstall and remove everything already done by bluefin image.

Finally, adds a section for F42


dreamyukii:
> epson inkjet packages are not available on F42

[the package is currently unmaintained](https://src.fedoraproject.org/rpms/epson-inkjet-printer-escpr)

castrojo (approved):

bpbeatty:
Hello,
May we please process the `EXCLUDED_PACKAGES` before the `INCLUDED_PACKAGES`?

Given the current workflow if one wants to install `vim-default-editor` in the all included packages and `nano-default-editor` in the all excluded packages the build fail with the error:
`  Problem: problem with installed package
    - installed package nano-default-editor-8.1-1.fc41.noarch conflicts with system-default-editor provided by vim-default-editor-2:9.1.737-1.fc41.noarch from fedora`

m2Giles:
Use dnf5 swap. We will look at changing order maybe in the future. However, dnf5 is much, much more aggressive on swapping packages.

Issue #2290, opened 2025-03-27
Move to terra switcheroo control


castrojo:
https://github.com/ublue-os/bluefin/blob/aeb37ff00243056e1d683e3fbda0e6881ff7ff0f/build_files/base/05-override-install.sh#L20

We might as well move to the terra package that aurora and bazzite are using since we need this for framework.

PR #2293, opened 2025-03-29
feat: enable terra and use switcheroo from there

inffy:
Enables terra repos and moves to use switcheroo-control from terra.

Taken from Aurora https://github.com/ublue-os/aurora/pull/221

Closes https://github.com/ublue-os/bluefin/issues/2290

### DIFF
--- a/build_files/base/05-override-install.sh
+++ b/build_files/base/05-override-install.sh
@@ -6,9 +6,9 @@ set -eoux pipefail
 
 if [[ "${UBLUE_IMAGE_TAG}" != "beta" ]]; then
     # Patched shells
-    dnf5 -y swap \
-        --repo=copr:copr.fedorainfracloud.org:ublue-os:staging \
-        gnome-shell gnome-shell
+   dnf5 -y swap \
+        --repo terra-extras \
+            gnome-shell gnome-shell
 
     # Fix for ID in fwupd
     dnf5 -y swap \
--- a/build_files/base/05-override-install.sh
+++ b/build_files/base/05-override-install.sh
@@ -14,15 +14,14 @@ if [[ "$(rpm -E %fedora)" -eq "40" ]]; then
         switcheroo-control switcheroo-control
 elif [[ "$(rpm -E %fedora)" -eq "41" ]]; then
     # Enable Terra repo (Extras does not exist on F40)
+    # shellcheck disable=SC2016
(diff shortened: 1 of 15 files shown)


castrojo (dismissed):

castrojo:
Dear future Bluefin maintainers:

We need switcheroo for Framework laptops, so taking a 3rd party dep here is not ideal but at least we know the fyra folks.

castrojo (dismissed):

build_files/dx/01-install-copr-repos-dx.sh
m2Giles, line 1: Do we need this enabled for -dx?

build_files/base/05-override-install.sh
```diff
@@ -6,21 +6,20 @@ set -eoux pipefail
 
 if [[ "${UBLUE_IMAGE_TAG}" != "beta" ]]; then
     # Patched shells
-    dnf5 -y swap \
-        --repo=copr:copr.fedorainfracloud.org:ublue-os:staging \
-        gnome-shell gnome-shell
+   dnf5 -y swap \
+        --repo terra-extras \
+            gnome-shell gnome-shell
```
m2Giles, line 11: We should probably configure a versionlock after the swap.

m2Giles (dismissed):

build_files/base/17-cleanup.sh
```diff
@@ -39,6 +39,10 @@ dnf5 -y copr disable ublue-os/staging
 dnf5 -y copr disable ublue-os/packages
 dnf5 -y copr disable che/nerd-fonts
 dnf5 -y copr disable phracek/PyCharm
+
+# Disable terra repos with dnf
+dnf5 -y config-manager setopt "terra*".enabled=false
```
m2Giles, line 44: This likely needs to be moved earlier or configured to ensure terra pkgs are not being pulled in when they shouldn't.

Builds are failing because of this.

m2Giles (changes requested):

build_files/base/17-cleanup.sh
inffy: Looking at what would be the best way. Maybe with setopt and then setting the terra priority to a high value?

m2Giles: I think I have a solution for you.

m2Giles (approved):

nothingneko:
Hey, here from Terra/Fyra, as is the same for Bazzite and Aurora please let us know if there's anything you need or things we can do to help make things better for you :3

PR #2297, opened 2025-03-31
chore: remove beta workarounds

m2Giles:
Mesa is now built for F42 in negativo17. Once merged into main this will work. Slowly removing other beta workarounds. Docker-CE is now building containerd.io but is still missing the rest of the stack.


bsherman:
I think we probably want this, just not yet... Just today we added more beta conditionals to Bluefin to get beta F42 builds going (working around rpmfusion), so once we do the F42 cutover, we could update this and then merge.

Or... just leave it, but it's probably better to remove it for the duration of the F42 life cycle to reduce maintenance complexity.

PR #2308, opened 2025-04-06
fix: remove old gtk3 themes

### LINKED ISSUE #2244: adw-gtk3 theme's flatpak is not maintained anymore
As described in the theme's official repo (issue [276](https://github.com/lassekongo83/adw-gtk3/issues/276)), the flatpak package is still on a very old version so the recommended install procedure has changed.

Here's the relevant part in the new readme: 

> 3. If you use flatpak applications it's recommended to use flatpak override. From a terminal run:
> ```bash
> sudo flatpak override --filesystem=xdg-data/themes
> ```
> 
> **Note:** This requires that the theme is installed in `~/.local/share/themes/`. Installing the theme as root, with a package manager, or any other method is not supported when it comes to flatpak.
> 
> To prevent outdated packages from flathub being installed, run: `sudo flatpak mask org.gtk.Gtk3theme.adw-gtk3 && sudo flatpak mask org.gtk.Gtk3theme.adw-gtk3-dark`
> 
> **IMPORTANT:** If you previously used the outdated flatpak packages, then uninstall them and prevent them from being installed with:
> ```bash
> flatpak uninstall org.gtk.Gtk3theme.adw-gtk3 org.gtk.Gtk3theme.adw-gtk3-dark && sudo flatpak mask org.gtk.Gtk3theme.adw-gtk3 && sudo flatpak mask org.gtk.Gtk3theme.adw-gtk3-dark
> ```

castrojo:
See ya!

Fixes #2244

### DIFF
--- a/bluefin_flatpaks/flatpaks
+++ b/bluefin_flatpaks/flatpaks
@@ -27,5 +27,3 @@ app/io.missioncenter.MissionCenter/x86_64/stable
 app/org.mozilla.firefox/x86_64/stable
 app/org.mozilla.Thunderbird/x86_64/stable
 app/org.gustavoperedo.FontDownloader/x86_64/stable
-runtime/org.gtk.Gtk3theme.adw-gtk3/x86_64/3.22
-runtime/org.gtk.Gtk3theme.adw-gtk3-dark/x86_64/3.22


castrojo:
Pryka is going to try to maintain this upstream so let's hold off for now.

Pryka:
> Pryka is going to try to maintain this upstream so let's hold off for now.

They add me as an maintainer, theme is updated now, should be soon available via flathub.

PR #2312, opened 2025-04-06
feat: add refine

castrojo:
Replaces GNOME Tweaks


bsherman (approved):
I just did a quick, non-exhaustive, comparison of GNOME Tweaks and Refine...

I noticed that Refine:
- adds ability to change Sound step value
- adds experimental features (VRR, Fractional Scaling, Xwayland native scaling)
- loses Sound theme option
- loses "Pointing Stick" options (eg, acceleration)
- loses distinct Touchpad options
- loses Keyboard option for "Overview Shortcut"
- loses Background image settings

I think the Background option is not a big deal, nor the Touchpad as GNOME settings has those handled pretty well... and I don't feel that Sound theme is significant here.

However, it seems Refine is maintained, but not Tweaks...

Overall, I think it's probably a better overall tool at this point, though that lack of changing "Overview Shortcut" or Pointing stick could be an issue for some users.

I'll approve, since supported software can be improved.

PR #2313, opened 2025-04-07
feat: add arm-image-installer to dx

lethedata:
Add arm-image-installer tools to dx.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -92,6 +92,7 @@
 			"dx": [
 				"adobe-source-code-pro-fonts",
 				"android-tools",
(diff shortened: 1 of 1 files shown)


lethedata:
I've had this layered for awhile and figured it might be useful to have in the base image as it's the defacto tool for Fedora IoT flashing.

castrojo:
Sorry but who is this for? This just appears to be a script you can copy into local/bin or something?

castrojo:
The android tools are there because lots of people use android. Sorry not keen on keeping a distro-specific package here.

PR #2325, opened 2025-04-09
fix: add build argument for RPMFusion mirror

tulilirockz:
Should fix some RPMFusion issues we've been getting with F42, added a build arg because im pretty damn sure we are going to run into this again at some point


m2Giles (approved):

Containerfile
m2Giles, line 1: You need to put this in bluefin's ARG block as well.

Right now it only gets passed on the -dx target step

m2Giles (changes requested):

Justfile
```diff
@@ -106,7 +106,7 @@ validate $image $tag $flavor:
 
 # Build Image
 [group('Image')]
-build $image="bluefin" $tag="latest" $flavor="main" rechunk="0" ghcr="0" pipeline="0" $kernel_pin="":
+build $image="bluefin" $tag="latest" $flavor="main" rechunk="0" ghcr="0" pipeline="0" $kernel_pin="" $rpmfusion_mirror="":
```
m2Giles, line 109: This also doesn't appear to be the right order.

Just takes argument position only. It doesn't do key-value pairing. Meaning in order for this to have an rpmfusion set. You need to make sure the arguments for rechunk, ghcr, pipeline, and kernel pin are set first.

m2Giles: Okay, this is fine. We have the wrapped recipes below. The issue was not adding the rpmfusion part to the recipe.

m2Giles:
we're nuking rpmfusion. So closing this.

PR #2346, opened 2025-04-14
fix(dx): install docker-ce from test repos in F42

p5:
While they aren't pushed to the stable repos

### DIFF
--- a/build_files/dx/04-override-install-dx.sh
+++ b/build_files/dx/04-override-install-dx.sh
@@ -28,4 +28,10 @@ tar --no-same-owner --no-same-permissions --no-overwrite-dir -xvzf /tmp/ls-iommu
 mv /tmp/ls-iommu/ls-iommu /usr/bin/
 rm -rf /tmp/ls-iommu*
 
+# Install Docker Release Candidate - remove this when F42 packages are pushed to the stable repos
+# https://download.docker.com/linux/fedora/42/x86_64/stable/Packages/
+if [[ $FEDORA_MAJOR_VERSION -eq 42 ]]; then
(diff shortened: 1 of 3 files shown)


castrojo (approved):

build_files/dx/04-override-install-dx.sh
```diff
@@ -28,4 +28,10 @@ tar --no-same-owner --no-same-permissions --no-overwrite-dir -xvzf /tmp/ls-iommu
 mv /tmp/ls-iommu/ls-iommu /usr/bin/
 rm -rf /tmp/ls-iommu*
 
+# Install Docker Release Candidate - remove this when F42 packages are pushed to the stable repos
+# https://download.docker.com/linux/fedora/42/x86_64/stable/Packages/
+if [[ $FEDORA_MAJOR_VERSION -eq 42 ]]; then
+    dnf install -y --enablerepo=docker-ce-testing docker-buildx-plugin docker-ce docker-ce-cli docker-compose-plugin
```
m2Giles, line 34: Please explicitly use dnf5. Also I think it's --enable-repo.

m2Giles (approved):

build_files/dx/04-override-install-dx.sh
p5: I'm out for a few hours so can't easily update this right nowsince this is only staying for about a week, can we push it through as-is, since it works

p5, line 34:
```suggestion
    dnf5 install -y --enablerepo=docker-ce-testing docker-buildx-plugin docker-ce docker-ce-cli docker-compose-plugin
```

PR #2355, opened 2025-04-16
fix(nvidia-open): correctly identify nvidia-open on image-info.json + fix live ISO nvidia

tulilirockz:
This writes `nvidia-open` instead of `nvidia` for the open kmod images, should work but I havent checked it yet.
Also makes it so the ISOs have the right container name there


tulilirockz:
This is absolutely working! 
image

PR #2378, opened 2025-04-20
fix(hwe): switch to skip-unavailable

m2Giles:
Iptsd doesn't appear to be available from sueface-linux right now.


p5:
The latest commit in the repo is adding F42 support, so maybe we wait it out?
Removing this may lead to a suboptimal experience

m2Giles:
We could add the skip-missing flag and it will auto pick up the package

tulilirockz (approved):

PR #2409, opened 2025-04-22
chore: Support Podman Machine in base images

m2Giles:
Rootful podman is necessary for certain privileged container actions. Often, these are run with --privileged and --security-opt label=disable effectively removing all security boundaries for the container.

Instead of doing that, we should have podman machine available on our images to encourage users who wish to use a rootful podman to instead use podman machine.

Once a default machine is created, podman --remote is transparently mapped to it and you can make rootful containers without affecting your host machine.


p5:
I would say being on DX is where this should be...  This also brings in a bunch of QEMU deps IIRC

m2Giles:
> I would say DX is the right place for this... It also brings in a bunch of QEMU deps IIRC

Podman already brings some of them in (the packaging is odd). It was in DX.

Issue #2403, opened 2025-04-21
stress-ng seems to have been omitted


JohnAtl:
### Describe the bug

ujust benchmark reports that the stress-ng command is not found

### What did you expect to happen?

The benchmark would run.

### Output of `bootc status`

```shell
❯ sudo bootc status
[sudo] password for john: 
● Booted ostree
           Commit: 4e00c4f023e4377b60f5b4d9fb2c23258fe09491ecf8fd1d3e80383014985203

  Rollback ostree
           Commit: 644454cb989a74e4ed3021d7bc46f1ad6a83693ef34394f08c3aef1f2bacfa0f
```

### Output of `groups`

```shell
❯ groups
john wheel docker
```

### Extra information or context

Also happens on the latest and stable isos.

castrojo:
The justfile needs to account for stressng being missing and install stress via brew instead.

PR #2413, opened 2025-04-24
chore: show a prompt for installing stress-ng via brew if its not installed

tulilirockz:
Fixes: https://github.com/ublue-os/bluefin/issues/2403


inffy:
didn't the stress-ng got removed and was replaced with `stress` from brew? But it requires the bluefin-cli as enabled that stress gets installed

PR #2422, opened 2025-04-26
fix: readd gnome-tweaks

castrojo:
need it for keyboard layouts. let's just bury it in the utilities folder


castrojo:
Don't merge maybe there's another flatpak tool.

bsherman (changes requested):
the spacing is inconsistent (even though I hate tab characters). it should be fixed

castrojo:
Ok I'm just going to add this for another cycle whatever.

PR #2427, opened 2025-04-27
feat: telnet client

lethedata:
Include telnet client in base image to allow controlling old or odd systems.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -71,6 +71,7 @@
 				"sssd-krb5",
 				"sssd-nfs-idmap",
 				"tailscale",
+				"telnet",
 				"tmux",
 				"ublue-bling",
 				"ublue-brew",


lethedata:
Telnet is available through inetutils in homebrew however that installs a lot of extra tools that already exist on the system. Telnet actually seems to be the only one missing.

castrojo:
Sorry we're not putting telnet on an image, having a few extra packages in brew isn't the end of the world.

lethedata:
> Sorry we're not putting telnet on an image, having a few extra packages in brew isn't the end of the world.

Just curious, what's the reasoning behind not including a telnet client?

castrojo:
We're not going to add things to images unless they are useful for the majority of users, telnet is easily installable through brew.

PR #2430, opened 2025-04-27
chore: add a few upstream bootc configuration files for nvidia and `bootc install`

tulilirockz:
This just adds a default filesystem type for Bluefin (btrfs) (makes it so bootc install to-disk works w/o modifications), and adds nvidia kargs on nvidia builds, shouldnt affect anything retroarctively

Fixes: https://github.com/ublue-os/titanoboa/issues/70


m2Giles:
Bad heredoc.

It's one line with no variables so use echo or printf

tulilirockz:
> Bad heredoc.
> 
> It's one line with no variables so use echo or printf

Oh its just that that TOML file needs the strings to be `"` characters, I thought a heredoc would be a bit easier to do instead of an `echo`

m2Giles:
If you wrapped in single quotes. Echo will work. Looks fine now

castrojo (approved):

PR #2433, opened 2025-04-28
feat(ci): switch to renovate from cronjobs

castrojo:
The intent here is to monitor silverblue-nvidia and then build right after instead of cron timing.

I've left gts and stable as weekly builds because they should just remain that way. LMK if I did this right. 🗡️


castrojo:
This should do until it's an action? I had nothing to do tonight. 🗡️

p5 (approved):

PR #2449, opened 2025-05-01
fix(just): add current user to dx-groups instead of root

vitalijusv:
A recent change introduced a regression, where the whole "ujust dx-group" script is running under pkexec, which means the $USER variable is now expanded "root" and not a current user.

To fix this, use just substitution to inject current user into the script.


castrojo (approved):

castrojo:
Thanks!

Issue #2436, opened 2025-04-28
Turn off automatic distrobox upgrades


castrojo:
This makes manual updates too slow and when people's distroboxes break they send us bug reports. These should be the user's responsibility not ours, let's just leave ourselves out of it.

Needs updating in Bluefin and Bluefin LTS.

MDutro:
I'll take this one on.

MDutro:
So... right away, I will have to admit to being a little confused. I have a machine on Bluefin:stable and `ujust update` does not attempt to run an upgrade on an Ubuntu distrobox container that I have.

Is this issue already resolved?

PR #2458, opened 2025-05-02
fix: remove Distrobox from automatic updates

MDutro:
This PR removes Distrobox container updates from the automatic background update service.

This PR addresses issue [https://github.com/ublue-os/bluefin/issues/2436](https://github.com/ublue-os/bluefin/issues/2436).

### DIFF
--- a/build_files/base/17-cleanup.sh
+++ b/build_files/base/17-cleanup.sh
@@ -4,6 +4,9 @@ echo "::group:: ===$(basename "$0")==="
 
 set -eoux pipefail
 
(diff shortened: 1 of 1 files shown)


MDutro:
I have added an if gate for Fedora 42

alexsch01:
Shouldn't this be >= 42? And not just an equals check

MDutro:
> Shouldn't this be >= 42? And not just an equals check

That was my initial thought. But discussion in the Discord led me to submit this conditional.

p5 (approved):

PR #2463, opened 2025-05-04
feat: use ublue-setup-services for setup services

tulilirockz:
This centralizes our setup service solution to the `ublue-setup-services` package, all these snippets came straight from Bluefin LTS.

I believe this should work fine, but it needs some testing, just to make sure everything is okay, although I am quite confident that this is fine because we've been using this for months on LTS
Fixes: https://github.com/ublue-os/packages/issues/339


tulilirockz:
This is a big PR but its meant to function exactly the same way as when we didnt move to these, anything that isnt working the same is a bug.

The units should be using `$USER/.local/share/ublue/setup_versioning.json` as their versioning file, its a little json thing that has all the script's versioning, instead of having that handling everywhere

castrojo (approved):

PR #2475, opened 2025-05-05
fix(just): use system gsettings to configure terminal transparency

akielbowicz:
This pull request updates the way `gsettings` is called within the `ptyxis-transparency` just recipe. I spent a significant amount of time over the weekend trying to achieve this, encountering some unexpected challenges.

Initially, I attempted to use `gsettings` and `dconf` (as suggested in [this gist](https://gist.github.com/andypiper/f75c25d91548af03fe586a19f35c9742)) to modify the opacity. However, these methods didn't seem to work as expected.

Ultimately, [this Stack Exchange answer](https://askubuntu.com/a/1007851) provided the necessary solution. I have another installation of `gsettings` (not sure what added it) which was being used instead of the system one.

Screenshot showing the Stack Exchange answer that provided the solution

p.s. Thanks for the hard work put into this project! It's been an amazing experience setting it up and a great way to learn a lot of new things.


castrojo (approved):
nice work! cc @ledif do you think we should unlink this in brew?

PR #2485, opened 2025-05-06
fix(just): bump rechunker and ISO generator

castrojo:


m2Giles (dismissed):

m2Giles:
I think removing bluefin-dx was a mistake.

Also added a renovate rule for digest pinning for rechunk and iso generator.

PR #2488, opened 2025-05-07
feat: use ublue-os-libvirt-workarounds for libvirt-related workarounds

tulilirockz:
Needs a PR on the packages repo right now


m2Giles:
I don't think these are needed since we rechunk

tulilirockz:
Pretty sure they are, we definitely need this on LTS at least, and thats rechunked

m2Giles:
The restorecon and the tmpfiles.d for creating swtpm's directory is needed. The mount over swtpm service shouldn't be needed anymore.

PR #2517, opened 2025-05-15
fix: create devmode version directory

bketelsen:
`/etc/ublue` doesn't exist, but the dx-groups service is trying to read a file from there, then write one later. This fix creates the directory.


castrojo:
cc @ledif @inffy @RealVishy

castrojo (approved):

m2Giles (approved):

RealVishy:
> cc @ledif @inffy @RealVishy 

https://github.com/ublue-os/aurora/pull/466

We put it in aurora, thanks!

PR #2526, opened 2025-05-15
fix: pin kernel to 6.14.5

castrojo:
We're getting reports that the gated kernel is failing with intel wifi cards, prepping this.

### DIFF
--- a/.github/workflows/build-image-stable.yml
+++ b/.github/workflows/build-image-stable.yml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         brand_name: ["bluefin"]
     with:
-#     kernel_pin: 6.13.8-200.fc41.x86_64 ## This is where kernels get pinned.
+      kernel_pin: 6.14.5-200.fc41.x86_64 ## This is where kernels get pinned.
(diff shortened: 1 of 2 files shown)


.github/workflows/build-image-stable.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       matrix:
         brand_name: ["bluefin"]
     with:
-#     kernel_pin: 6.13.8-200.fc41.x86_64 ## This is where kernels get pinned.
+      kernel_pin: 6.14.5-200.fc41.x86_64 ## This is where kernels get pinned.
```
p5, line 25: Think this will need to be a F42 version.  Rolling back from 42 to 41 isn't something we can (easily) do.

p5 (approved):
Will approve so you can merge if necessary.
The changes LGTM

bsherman (approved):

castrojo:
Will this work? looking at the failing PRs, .5 never hit coreos so it's not cached.

bsherman:
Ugh no. It won't work in that case

castrojo:
OK we'll just tell people to go to latest if they have this.

Issue #1526, opened 2024-07-22
Add explicit fedora version to channel chooser widget


castrojo:
![image](https://github.com/user-attachments/assets/166ba83c-943d-41d6-8b90-bcab928ecd1c)

It'd be nice if we could show the F40, F41, etc versions of Fedora for people who want to lock into a version instead of a rolling tag.

dnikoloski:
Do we want this hard-coded like:
Choose:
```
  date (F40)
  latest (F40)
  gts (F39)
  stable (F40)
```

castrojo:
Yeah but ideally we don't want to have to update each one manually every time if there's a release if possible. 

See also: https://github.com/ublue-os/config/issues/251

Malix-off:
Would be nice to also think about re-ordering the tags like :

> | Name | Fedora Version | Description |
> | - | - | - |
> | _the currently selected one (remove the associated element from the list bellow)_ | _associated version_ | _associated description_ |
> | gts | 40 | Default |
> | stable | 41 | Enthusiasts |
> | Latest | 41 | Testers |
> | Date | Fixed Version | Fixed date release |

AtomHare:
With the new PR there's now fedora versions dynamically fetched shown as `latest (F42)`, `gts (F41)` etc

castrojo:
Thanks for this Atom!

PR #2530, opened 2025-05-16
feat: Add fedora version to the rebase helper

AtomHare:
This PR adds fedora versions next to the rebase tag. Those are not hardcoded and use `skopeo` to extract the version.

Then jq is used to extract `org.opencontainers.image.version`, which is usually `gts-41.xx`, `latest-42.xx` or `42.xx` for stable and stable-daily. That's why `sed` is used to match everything before the `-`. `cut` is then used to only retrieve the 41/42 part.

`choose_target=${choose_target%% *}` allows matching everything after the first space and remove it.

Related to #1526

### DIFF
--- a/system_files/shared/usr/bin/ublue-rollback-helper
+++ b/system_files/shared/usr/bin/ublue-rollback-helper
@@ -22,9 +22,9 @@ function list_tags(){
 
 function rebase_helper(){
     base_image="${IMAGE_REGISTRY}/${IMAGE_NAME}"
-    echo "Which Tag would you like to rebase to?"  
+    echo "Which Tag would you like to rebase to?"
 
-    if [[ "$composefs_enabled" == "true" ]]; then        
+    if [[ "$composefs_enabled" == "true" ]]; then
         CHANNELS=(latest stable stable-daily)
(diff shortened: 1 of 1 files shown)


m2Giles:
This appears to remove the convenience tags?

Additionally, this is a lot of skopeo calls. stable/stable-daily do not have an associated Fedora version so those should not be inspected. GTS and latest are always offset by 1. So you could inspect 1 and infer the other.

AtomHare:
What do you mean by convenience tag ?

About inferring. how would it work close to a new fedora release like F43 is gts and latest still 1 or 2 versions apart.

If this is not a problem I could query latest (gts = latest - 1) and only stable then ?

m2Giles:
Okay I see it only adds it as (FXX) after the stream name.

You would need to inspect 2 of the images. Stable and one of latest/GTS.

With GTS or Latest you can get the version of the other through math. They should never be on the same version. Stable can be on either one (GTS and stable effectively match for about a month each cycle)

AtomHare:
I updated with your recommendations, it should be 100% faster when fetching `(latest stable stable-daily gts)` and 50% faster when fetching `(latest stable stable-daily)` since it will only retrieve `latest` and `stable` at most.

I also rebased the PR hence the force push

PR #2540, opened 2025-05-17
fix: remove terra repos and overrides

castrojo:


tulilirockz:
I think you need to remove the part that disable the repos too

tulilirockz (approved):
LGTM: we can figure out switcheroo later

PR #2574, opened 2025-05-24
fix(pin): linux-firmware to 20250410

travankor:
This PR fixes a major regression with Atheros firmware, where the current firmware prevents the kernel from loading the module.

https://bugzilla.kernel.org/show_bug.cgi?id=220108
https://bugzilla.redhat.com/show_bug.cgi?id=2365882
https://www.reddit.com/r/archlinux/comments/1kja6f9/ath12k_regression_on_latest_linuxfirmware_upgrade/

### DIFF
--- a/build_files/base/18-workarounds.sh
+++ b/build_files/base/18-workarounds.sh
@@ -14,10 +14,11 @@ set -eoux pipefail
 #    rpm-ostree override replace https://bodhi.fedoraproject.org/updates/FEDORA-2024-dd2e9fb225
 #fi
 
-if [ "$FEDORA_MAJOR_VERSION" -eq "42" ]; then
-# Workaround atheros-firmware regression, see https://bugzilla.redhat.com/show_bug.cgi?id=2365882
-    rpm-ostree override replace https://bodhi.fedoraproject.org/updates/FEDORA-2025-0b94571b06
-fi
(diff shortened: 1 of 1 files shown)


renner0e:
We might want to use dnf for that.
Quick local testing shows that this seems to work.

We should need this for GTS as well right? So greater and equals F41 for the if statement 
```
if [ "$FEDORA_MAJOR_VERSION" -ge "41" ]; then
# Workaround atheros-firmware regression, see https://bugzilla.redhat.com/show_bug.cgi?id=2365882
    dnf -y swap atheros-firmware atheros-firmware-20250410-1$(rpm -E %{dist})
fi
```

But I don't know if that causes problems because every other firmware package would be newer than `atheros-firmware`


For future reference:

`dnf list --showduplicates atheros-firmware`
```
Available packages
atheros-firmware.noarch 20250311-1.fc42 fedora
atheros-firmware.noarch 20250410-1.fc42 updates-archive
atheros-firmware.noarch 20250509-1.fc42 updates-archive
atheros-firmware.noarch 20250509-1.fc42 updates
```

castrojo (approved):

castrojo:
I rolled out a new fresh stable tag earlier today so we can ship this asap, thanks!

PR #2594, opened 2025-05-31
fix: pin mutter to .48.1-1 in F42

castrojo:
Avoids https://bugzilla.redhat.com/show_bug.cgi?id=2369147

### DIFF
--- a/build_files/base/18-workarounds.sh
+++ b/build_files/base/18-workarounds.sh
@@ -14,6 +14,13 @@ set -eoux pipefail
 #    rpm-ostree override replace https://bodhi.fedoraproject.org/updates/FEDORA-2024-dd2e9fb225
 #fi
 
+# Workaround mutter regression, remove in early June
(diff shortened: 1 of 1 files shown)


DarkGhostHunter:
What's the difference between Mutter `48.2-2-f42` and `48.1-1-f42`? Seems that both are bug free.

castrojo:
Ok I had to disable v4l2 to get it to build, looks like these are going to pass so we'll publish these as the weeklies tonight. 

(The version # is the one that was recommended in the bugzilla ticket)

DarkGhostHunter:
> Ok I had to disable v4l2 to get it to build, looks like these are going to pass so we'll publish these as the weeklies tonight.
> 
> (The version # is the one that was recommended in the bugzilla ticket)

Yeah, seems everything is fine with 48.1-1. I also have some cursor bugs on 48.2, so it seems like the safest bet until it's resolved.

PR #2606, opened 2025-06-02
fix(changelogs): Remove Devpod entry now that it's a flatpak

AtomHare:
Devpod is now a flatpak this remove the following entry from changelogs since it wasn't populated properly (because it's not a package anymore)


p5 (approved):
LGTM!  Thanks!

PR #2676, opened 2025-06-20
fix(bazaar): update firefox entry

castrojo:


castrojo:
Moving pika to utilities since shipping 2 backups is confusing.

PR #2690, opened 2025-06-22
feat: remove gnome-software

castrojo:
Bazaar is already landed, removing this.


castrojo:
Do not merge until: https://github.com/ublue-os/packages/pull/638

And the corresponding schema bump

PR #2706, opened 2025-06-26
Kernel pin

castrojo:
Reported in chat: https://discussion.fedoraproject.org/t/system-is-unstable-after-kernel-update-to-6-15-3-200-fc42/156416

### DIFF
--- a/.github/workflows/build-image-gts.yml
+++ b/.github/workflows/build-image-gts.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         brand_name: [bluefin]
     with:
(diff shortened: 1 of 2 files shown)


inffy:
Think we have sme leeway here.

stable is on 6.14.9 (coreOS) and looks like the next coreOS will be on 6.14.11 which is the last 6.14 version.

So I guess after two weeks we could pin stable the 6.14. Although not sure what version coreOS will move after that and what is the status with 6.15 at that point.

castrojo:
Closing this then, thanks for the detective work!

PR #2727, opened 2025-07-03
chore: Remove Name key from [Unit] ublue-fix-hostname.service

Yunraze:
I was doing a `systemd-analyze` as a part of my integrity checks to my bluefin, and noticed that `ublue-fix-hostname.service` has an unknown key `Name` in the `[Unit]` section. This is a super minuscule thing, but from a correctness standpoint there shouldn't be any keys that just get ignored. `Name` isn't a part of the `[Unit]` as per https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#%5BUnit%5D%20Section%20Options

### DIFF
--- a/system_files/shared/usr/lib/systemd/system/ublue-fix-hostname.service
+++ b/system_files/shared/usr/lib/systemd/system/ublue-fix-hostname.service
@@ -1,5 +1,4 @@
 [Unit]
-Name=Universal Blue Missing /etc/hostname Workaround
 Description=Workaround for the missing /etc/hostname file on Universal Blue systems
 After=network.target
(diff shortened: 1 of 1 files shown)


Yunraze:
Some checks errors, but seemingly not related to my change:

```
 + '[' 575.64-1.fc41 '!=' 575.64.03-1.fc41 ']'
  + echo 'Error: kmod-nvidia version (575.64-1.fc41) does not match nvidia-driver version (575.64.03-1.fc41)'
  Error: kmod-nvidia version (575.64-1.fc41) does not match nvidia-driver version (575.64.03-1.fc41)
```

PR #2777, opened 2025-07-17
chore: pin stable to kernel 6.14.11

inffy:
Next CoreOS seems to move to 6.15.4 kernel which is affected by the btrfs bug. We should pin stable until a fixed kernel version is available (looks like 6.15.6 is still affected)


castrojo:
We'll need to do gts too ...

PR #2824, opened 2025-07-24
42: Add sudo-rs as the default sudo

endocrimes:
Copying ubuntu, bazzite (https://github.com/ublue-os/bazzite/pull/2954), and aurora (https://github.com/ublue-os/aurora/pull/728), switch to sudo-rs by default on Bluefin:42.

Not sure if `cleanup` is the right place to bind these, but it's also not... super obvious where else (base-image-changes?).

(currently untested - just opening for discussion)


castrojo:
Ubuntu is committed to shipping this in the fall around the time F43 ships, I think that'd be a good time to switch but then that'll leave old-sudo in GTS, which is probably a good idea anyway.

castrojo:
bluefin-cli opts into all the rust stuff including uutils so you can live the dream today, other than sudo-rs. 😒  

`sudo` should be on the image though which is why we wouldn't do it in brew, just in case anyone wants to burn me at the stake.

tulilirockz:
Lets just test this a bit before changing, I personally feel like this 100000000% should be the default though

castrojo:
Perhaps someday a new species will just ... be that way ...

ohno

castrojo:
We'll do this in other bluefins and just leave this one how it is.

PR #2949, opened 2025-08-07
chore: change default filesystem from btrfs to xfs

hanthor:

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,8 +1,8 @@
 blank_issues_enabled: true
 contact_links:
   - name: Ask Bluefin
-    url: https://github.com/ublue-os/bluefin/discussions/categories/ask-bluefin
+    url: https://ask.projectbluefin.io
     about: Ask our trained raptor questions about Bluefin
   - name: Questions and Feedback
-    url: https://universal-blue.discourse.group/c/bluefin/6
+    url: https://community.projectbluefin.io
     about: Share tips and help others
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -88,7 +88,7 @@ jobs:
       - name: Setup Syft
         id: setup-syft
         if: false
-        uses: anchore/sbom-action/download-syft@7b36ad622f042cab6f59a75c2ac24ccb256e9b45 # v0
+        uses: anchore/sbom-action/download-syft@da167eac915b4e86f08b264dbdbc867b61be6f0c # v0
         with:
           syft-version: v1.20.0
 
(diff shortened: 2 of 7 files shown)


castrojo:
Let's ensure we don't spin any ISOs except to GH artifacts to test, then we can push to cloudflare after some testing. Then we can probably freshen them up at the same time LTS releases.

castrojo:
You don't need to manually merge in main every time the merge queue would handle that. Ok is this ready?

ahmedadan:
> You don't need to manually merge in main every time the merge queue would handle that. Ok is this ready?

Old habits! This is ready yes!

hanthor (commented):
This looks ready

ahmedadan:
Looks like @castrojo fixed the font failure on main, can we kick off the CI again?

ahmedadan:
Do we need something else done before we can merge?

castrojo:
We'll need to test this so not keen until closer to F43, we need to finish off LTS and fix bugs before moving on to this. We'll just keep it here queued up.

hanthor:
A good thing to test would be rebasing to LTS from a XFS fedora install. We maybe need to edit the EFI dir. But we might want that to be bluefin anyway, instead of Fedora

alexsch01:
Isn't btrfs superior to XFS for a desktop Linux user (and not server)?

There are way more bluefin users then bluefin-lts users, so it doesn't make sense to me to change it just for the LTS rebasing use case that most people wouldn't do

hanthor:
Also btrfs recently had regressions in the kernel. RHEL doesn't support for some good reason, etc.

Are really this many people using subvolumes? Btrfs isn't going away its just going to be opt-in. People who don't care, won't care and just roll with default XFS. We could to ext4 instead???

alexsch01:
> Are really this many people using subvolumes? Btrfs isn't going away its just going to be opt-in. People who don't care, won't care and just roll with default XFS. We could to ext4 instead???

I thought btrfs was superior for snapshots
There was a reason Fedora uses it by default

pedrohqb:
You could easily opt in for btrfs on Anaconda or would you have to resort to manual installation?

castrojo:
Yeah you'd choose manual partitioning and choose btrfs. Can we do some test ISOs to test ahead of time?

castrojo:
We don't have time to test this thoroughly so punting until next cycle.

pedrohqb:
Even though now closed, I would like to add my 2 cents on the matter: I would rather not change as that will add more maintenance burden on Bluefin devs since, as it is now, the stability is somewhat guaranteed by the Fedora upstream. If it that change does not not add anything "new" regarding functionality - I only see the possibility of moving between Bluefin LTS and Bluefin (GTS) -, I would keep as it is. If it is not broken, do not fix it...

PR #2961, opened 2025-08-08
fix(ISO-workflow): fix logic to prevent LTS nvidia-open job

hanthor:
adding the image version options broke the original logic. moved it to a if statement


castrojo (approved):

castrojo:
This seems ready to go?

hanthor:
Nah it isn't quite right. I'll have to take another look

castrojo:
We fixed this with the new workflow!

PR #2967, opened 2025-08-10
revert: remove gradia and the emoji picker.

castrojo:
Removing since the keyboard shortcuts don't work quite right.


hanthor:
We probably need to add a script to add the shortcuts if they don't exist? I haven't gotten to test this out yet

castrojo:
We don't have time for things like this we have way larger issues to fix, we can pause on the features for a while.

Issue #2842, opened 2025-07-26
Internal audio controller on Framework 13 AMD Ryzen HX370 no longer recognised


ported-pw:
### Describe the bug

After updating to [gts-20250720](https://github.com/ublue-os/bluefin/releases/tag/gts-20250720) from [gts-20250706](https://github.com/ublue-os/bluefin/releases/tag/gts-20250706), the internal audio controller and thus the 3.5mm headphone jack on the Framework 13 HX370 no longer works.

The card does not show up under pactl, the issue is already earlier:
```
bash-5.2# lspci -k | grep -A 3 "c1:00.6"
pcilib: Error reading /sys/bus/pci/devices/0000:00:08.3/label: Operation not permitted
c1:00.6 Audio device: Advanced Micro Devices, Inc. [AMD] Family 17h/19h/1ah HD Audio Controller
    Subsystem: Framework Computer Inc. Device 000b
    Kernel modules: snd_hda_intel
c2:00.0 Non-Essential Instrumentation [1300]: Advanced Micro Devices, Inc. [AMD] Strix/Strix Halo PCIe Dummy Function
bash-5.2# ls -la /sys/bus/pci/devices/0000:c1:00.6/driver
ls: cannot access '/sys/bus/pci/devices/0000:c1:00.6/driver': No such file or directory
bash-5.2# echo "0000:c1:00.6" | sudo tee /sys/bus/pci/drivers/snd_hda_intel/bind
0000:c1:00.6
tee: /sys/bus/pci/drivers/snd_hda_intel/bind: Device or resource busy
bash-5.2#
```
The driver does not get loaded, and by default the device is disabled and has an invalid IRQ:
```
❯ sudo lspci -s c1:00.6 -v
c1:00.6 Audio device: Advanced Micro Devices, Inc. [AMD] Family 17h/19h/1ah HD Audio Controller
    Subsystem: Framework Computer Inc. Device 000b
    Flags: fast devsel, IRQ 255, IOMMU group 29
    Memory at b04c0000 (32-bit, non-prefetchable) [disabled] [size=32K]
    Capabilities: [48] Vendor Specific Information: Len=08 <?>
    Capabilities: [50] Power Management version 3
    Capabilities: [64] Express Endpoint, IntMsgNum 0
    Capabilities: [a0] MSI: Enable- Count=1/1 Maskable- 64bit+
    Capabilities: [100] Vendor Specific Information: ID=0001 Rev=1 Len=010 <?>
[...]

PR #2972, opened 2025-08-10
fix: update logic for applying framework laptop 13 fixes

kreed:
Two changes:
1. The suspend workaround is specific to Framework 13 Ryzen 7040 but we're also applying it to the recently released Ryzen AI 300 models. These don't need the workaround.
2. The 3.5mm jack workaround is needed for all Framework 13 models. The kernel code applies this workaround automatically for most current Intel/AMD framework models, but it's missing the new Ryzen AI 300. It doesn't seem to hurt to apply this workaround explicitly even when the kernel does it (we've been doing so for the Ryzen 7040 for some time now) so I loosened the condition to apply it to all models.

I've tested this on a Framework 13 Ryzen AI 300. If anyone could test on other Framework 13 models (especially the Ryzen 7040) that would be appreciated!

### DIFF
--- a/system_files/shared/usr/share/ublue-os/system-setup.hooks.d/10-framework.sh
+++ b/system_files/shared/usr/share/ublue-os/system-setup.hooks.d/10-framework.sh
@@ -40,32 +40,35 @@ fi
 
 SYS_ID="$(cat /sys/devices/virtual/dmi/id/product_name)"
 
-# FRAMEWORK 13 AMD FIXES
-if [[ "$VEN_ID" == "Framework" && "$SYS_ID" == "Laptop 13 ("* && "$CPU_VENDOR" == "AuthenticAMD" ]]; then
-    echo "Framework Laptop 13 AMD detected"
+# FRAMEWORK 13 FIXES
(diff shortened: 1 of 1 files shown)


ledif (commented):
Thanks for these changes! I tested with my Framework 13 7840U and it properly detected the machine and entered the correct branches.

We will need to port this to Aurora as well.

kreed:
I added another commit that removes the modprobe flag for the headset jack. Background:

1. The kernel level quirk is updated to detect AI 300 Framework 13's for 6.17. It's also in the stable releases (6.16.12, 6.15.11, and the LTS kernels too). So when using a new enough kernel the quirk should be automatically applied for all Framework 13 models.
2. Did some reading on the framework forums and it seems the kernel-level quirk is actually slightly better than the modprobe one (dell-headset-multi can result in duplicate inputs showing up). https://community.frame.work/t/request-kernel-patch-for-the-dell-headset-multi-workaround/18222/6
3. #2842 reported that removing the modprobe fixed the issue

Given all this, I think it makes sense to remove it. May want to hold off on merging this change until all the releases are using a kernel which includes the fix though.

hanthor (approved):
Can't test but if it works for @ledif it should work for us

PR #3005, opened 2025-08-15
fix: remove ramalama from image

hanthor:
we're going to the use the brew package. it keeps getting out of date, let's cut out the distro


castrojo:
Let's roll this one out after the weeklies on saturday, I'll land it.

castrojo (approved):

PR #3035, opened 2025-08-21
feat: add ramalama installed from brew on fist login after DX

hanthor:


castrojo:
I don't understand why this is a service? Why not just brew install it when the user runs dx mode like we do with the flatpaks?

castrojo:
We could put it here and install the brewfile for dx right after the flatpaks: https://github.com/ublue-os/bluefin/blob/de0bc81f364311186ba127181ee78e9f17fabef1/just/bluefin-system.just#L43

Issue #2905, opened 2025-08-01
Add bbrew to the motd


jfmongrain:
Now that we ship bbrew, it would be a good idea to add it to the command list in the motd, under the "brew help" entry, or maybe even replacing it.

castrojo:
Yeah can someone mockup what it would look like? We have new tips coming too.

jfmongrain:
I mean, let's keep it simple?

![Image](https://github.com/user-attachments/assets/caa17e3d-3694-4bc5-b86e-b034a1363423)

castrojo:
We have room on there, how about:

`brew help` or `bbrew` 

To show off both?

jfmongrain:
Good idea.

victoitor:
Is it being shipped in gts? Installed it last night and bbrew is not present. The message becomes weird.

castrojo:
You need to install bbrew, what error do you get?

victoitor:
The error is of "Command not found". It's not a big issue, but it's odd to mention it in motd to use a command which is not installed.

Motd shows up like this: 

<img width="997" height="200" alt="Image" src="https://github.com/user-attachments/assets/c29b889d-86c3-46ea-a731-0dbb914d9f09" />

Installed gts and upgraded to bluefin-dx:gts. Found about this command when I was reading the [Command Line documentation](https://docs.projectbluefin.io/command-line/).

As a recommended package both in modt and in the docs, I did not find it easy to install as `brew search bbrew` doesn't find it.

castrojo:
Yeah I backed it out of the MOTD, it's part of the `bluefin-cli` thing so you have to turn that off and back on. I've removed the message in the next build.

victoitor:
Removing and adding it again worked.

PR #3042, opened 2025-08-22
Feat: Add bbrew to the motd (#2905)

mheuvel-dev:
Adds `bbrew` to motd according to the suggestion by @castrojo in #2905


castrojo (approved):
Nice! Thanks!

PR #3078, opened 2025-08-28
Add Bazaar to LTS ISOs

hanthor:


hanthor:
looks like it's adding Bazaar for LTS and not for non-LTS

castrojo (approved):

Issue #3090, opened 2025-08-29
motd unnecessary slash in url & tldr issue


jiesou:
### Describe the bug

Here are 2 small motd issue

## slash

<img width="1187" height="860" alt="Image" src="https://github.com/user-attachments/assets/8bbfd008-2b4c-40e7-b77b-15582af61169" />

https://github.com/ublue-os/bluefin/blob/ed56af9a8320a5d8d1a1d934740db67686e94d30/system_files/shared/usr/share/ublue-os/motd/template.md?plain=1#L14-L17

This is not a big deal, but when you notice it, the slash at the end of the URL is very ugly.

## tldr tips

https://github.com/ublue-os/bluefin/blob/4dd4f4d11c2c5401125b86218841e8aaf23cb972/system_files/shared/usr/share/ublue-os/motd/tips/10-tips.md?plain=1#L10

As far as I know, tldr is not installed by default without bluefin-cli. So this may cause some confusion.

### What did you expect to happen?

remove slashs and tldr tips

### Output of `bootc status`

```shell
● Booted ostree
           Commit: 8ce389f9bbf6e37b46962235900536d12a2bf038528e8085d02a4ba8fbd6c052

  Rollback ostree
           Commit: e9e7051d139540f489b35734be587d3d0c4dbc84bded0148c5313ef331b53552
```

### Output of `groups`

```shell

```

PR #3096, opened 2025-08-29
fix: motd unnecessary slash in url (#3090)

mheuvel-dev:
Makes the slashes in motd urls consistent, see https://github.com/ublue-os/bluefin/issues/3090


p5 (approved):
Thanks!

castrojo (approved):

PR #3299, opened 2025-09-30
chore: remove bcache-tools

renner0e:
picked from:
https://github.com/ublue-os/aurora/pull/1029


castrojo:
I agree we can remove this, but let's do it during the F43 move. This doesn't have anything to do with the bcache filesystem, this is tiered caching using any fs. Though these days I'm not sure it's worth caring about spinning rust.

PR #3307, opened 2025-10-02
feat(bazaar): use light/dark variant images + jxl + flatpak path

renner0e:
You can safely merge this whenever you feel like it,, even before the beta as it is not destructive to the behavior of the current setup.

This change will only be used by the flatpak, so instead of replacing our existing config, I added them in the correct place (/etc) for the flatpak. This makes it so rpm and flatpak can coexist and the flatpak will use the new config.

~~The last 2 commits also adjust the path to be used with the flatpak.
Merge this in tandem with: https://github.com/ublue-os/bluefin/pull/3352~~

this makes it so that when you toggle light/dark mode bazaar will have
the respective day/night wallpaper + gradient

Also happens to save a little bit of space because jxl

## old look:

https://github.com/user-attachments/assets/bc2e3c74-2403-4804-9810-181e40ed3480

## new look:

https://github.com/user-attachments/assets/97519564-3b00-4657-8e7b-d34f2f495b61

## new look but constantly switching between light/dark because why the hell not:

https://github.com/user-attachments/assets/38d166f2-3e92-4b11-b8b6-254c7bfaf75a

image
image

### DIFF
--- a/system_files/shared/etc/bazaar/config.yaml
+++ b/system_files/shared/etc/bazaar/config.yaml
@@ -125,8 +125,8 @@ sections:
   - title: "Bluefin Recommends"
     subtitle: "Our Favorite Applications"
     rows: 3
-    light-banner: file:///usr/share/ublue-os/bazaar/11-bluefin-day.jxl
-    dark-banner: file:///usr/share/ublue-os/bazaar/11-bluefin-night.jxl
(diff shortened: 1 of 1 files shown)


castrojo:
This needs an updated bazaar or should be merged after the switch to the flatpak.

renner0e:
> This needs an updated bazaar or should be merged after the switch to the flatpak.

just needs the new version, how we get it doesn't matter

PR #3326, opened 2025-10-05
fix: replace unofficial tailscale gui with official option

castrojo:
Removed Tailscale extension checks for Silverblue variant. We're using the new feature in the upstream app itself


p5 (approved):
LGTM - assuming the new feature is already being shipped in Bluefin

castrojo:
Let's land this for F43 so we can add it to the release notes with other small changes.

PR #3327, opened 2025-10-05
feat: remove old packages

castrojo:
This removes all those fonts and the F41/F42 specific sections. Adds uupd to the main list.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -5,6 +5,7 @@
 				"adcli",
 				"adw-gtk3-theme",
 				"bash-color-prompt",
+				"bazaar",
 				"bcache-tools",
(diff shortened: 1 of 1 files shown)


hanthor:
Are we switching to Bazaar flatpak?

skyler544:
This isn't going to remove `evolution-ews-core` when gts switches to fedora 42, right? Just noting this here because that package was recently added and I love it when I can remove things from my downstream build 🧐

https://github.com/ublue-os/bluefin/commit/805ef67b5ae37ad99b1e8e4aea7ed9700e6c5c7e

castrojo:
We redid this anyway, ews is in there for all upcoming releases.

Issue #3269, opened 2025-09-25
Add framework_tool on framework laptops


castrojo:
Let's use the detect logic for the logomenu to install framework_tool from ublue-os/homebrew-tap in the background.

PR #3328, opened 2025-10-05
feat: add framework-tool on framework laptops

ahmedadan:
closes: #3269

### DIFF
--- a/system_files/shared/usr/share/ublue-os/system-setup.hooks.d/10-framework.sh
+++ b/system_files/shared/usr/share/ublue-os/system-setup.hooks.d/10-framework.sh
@@ -69,3 +69,17 @@ if [[ "$VEN_ID" == "Framework" && "$SYS_ID" == "Laptop 13 ("* && "$CPU_VENDOR" =
         fi
     fi
 fi
+
+# Install framework_tool for Framework laptops
+if [[ ":Framework:" =~ :$VEN_ID: ]]; then
+    echo "Framework laptop detected, installing framework_tool in the background"
+
+    # Check if Homebrew is available and if it is tap and install framework_tool
+    if command -v brew &> /dev/null; then
+        echo "Installing framework_tool from ublue-os/homebrew-tap"
+        brew tap ublue-os/tap
+        brew install --cask framework_tool
+    else
+        echo "Warning: brew not found, skipping framework_tool installation"
+    fi
+fi


system_files/shared/usr/share/ublue-os/system-setup.hooks.d/10-framework.sh
```diff
@@ -40,34 +40,46 @@ fi
...
+
+# Install framework_tool for Framework laptops
+if [[ ":Framework:" =~ :$VEN_ID: ]]; then
+    echo "Framework laptop detected, installing framework_tool in the background"
+
+    # Check if Homebrew is available and if it is tap and install framework_tool
+    if command -v brew &> /dev/null; then
+        echo "Installing framework_tool from ublue-os/homebrew-tap"
+        brew tap ublue-os/tap
+        brew install --cask framework_tool
```
castrojo: Shouldn't one offs be `brew install --cask ublue-os/tap/framework_tool`? 

Not sure of the ramifications.

castrojo: Also what happens if I'm not on the net, will it try to install this next time?

ahmedadan: > Shouldn't one offs be `brew install --cask ublue-os/tap/framework_tool`?
> 
> Not sure of the ramifications.

We can move to the one off, both tap and then install so no difference between them as far as functionality.


> Also what happens if I'm not on the net, will it try to install this next time?

That's beyond my expertise, when would the next time this hook get called?

castrojo: Not sure, I think we just haven't been in this part of the codebase in a while, we should perhaps take a look at the logic as a whole. 

We'd want to install the wallpapers in this section probably too? Also should check with daeglus on production status of the tap before we add it here. This is gonna be so good!

hanthor: We can write a file framework-tool.installed and check for it so if it exists we know it was installed at some point. That way users can remove it without it keep coming back, in case we want that

ahmedadan: > Not sure, I think we just haven't been in this part of the codebase in a while, we should perhaps take a look at the logic as a whole.

I'll take a closer look and see if I can understand the flow a little better.

> We'd want to install the wallpapers in this section probably too? Also should check with daeglus on production status of the tap before we add it here. This is gonna be so good!

Great idea! I'll connect with daeglus and report back on the status. I also don't have permissions in this repo, can one of you mark this PR as a draft?

> We can write a file framework-tool.installed and check for it so if it exists we know it was installed at some point. That way users can remove it without it keep coming back, in case we want that

Do we have a standardized locations for these types of files?

ahmedadan:
Addressed the comments and made some assumptions, would love some extra eyes. @castrojo @hanthor

hanthor:
I like the idea of having one script per vendor to add tools and wallpapers or whatever

system_files/shared/usr/share/ublue-os/system-setup.hooks.d/10-framework.sh
tulilirockz, line 5:
```suggestion
version-script framework system 3 || exit 0
```
Lets also bump the release so that people get these new ones

system_files/shared/usr/share/ublue-os/system-setup.hooks.d/10-framework.sh
```diff
@@ -71,3 +71,40 @@ if [[ "$VEN_ID" == "Framework" && "$SYS_ID" == "Laptop 13 ("* ]]; then
...
+            else
+                echo "Warning: framework-wallpapers installation failed, will retry on next run"
+            fi
+        else
+            echo "Framework wallpapers already installed, skipping"
+        fi
+    else
+        echo "Warning: brew not found, skipping Framework software installation (will retry when brew is available)"
+    fi
+fi
```
tulilirockz: We need to move this to the `user-setup.hooks.d` instead of this file, as a new script (probably just 10-framework.sh still), so that if this fails it wont break the system stuff. This script runs as root and it will break homebrew

tulilirockz (changes requested):
Lets just fix those things I commented on first

tulilirockz:
Should we move this to a Just recipe? As much as I dont like them, I feel like that would be more fitting. Grandma aint touching framework_tool ever

ahmedadan:
@castrojo @tulilirockz Is this PR something we still want?

castrojo:
Yes we still need this!

hanthor (approved):

hanthor (approved):

castrojo (approved):

PR #3330, opened 2025-10-05
cleanup: remove a bunch of fonts

castrojo:
TODO: add the ubuntu and inter packages back via packages.


castrojo:
Ok I don't think we need inter since adwaita is on the image already. The ubuntu font we can probably bring in via brew in an installation step if we care that much.

hanthor:
Yeah if we switched to adwaita this should be good to go

hanthor (approved):

PR #3331, opened 2025-10-05
fix(iso): disable suspend during installation and initial setup

### LINKED ISSUE #2590: Disable suspend during installation
### Describe the bug

my computer was not able to recover from the suspend. I got an empty screen that displayed keyboard-input and no disc-activity at all. ctrl-alt-del gave a list of systemd-failures.

I expect that both during installation and at user-account setup suspend is completely disabled! even a laptop can be trusted to cope with that little battery-loss. of course once everything finished, suspend during login-screen and after login should be enabled by default for the user to disable it as needed.

### Output of `bootc status`

```shell
❯ sudo bootc status
● Booted image: ghcr.io/ublue-os/bluefin:gts
        Digest: sha256:643e2b082a4d6a2cab6b799e13776fb013b513b6793d01cc4bb8f9b4563df4ac (amd64)
       Version: gts-41.20250518 (2025-05-18T06:01:46Z)

~ took 2m17s
```

### Output of `groups`

```shell

```

actually during installation I disabled power-saving. however at first boot-up I am asked to create an user-account. during this process screen-saving and automatic suspend is enabled. with a slow disc or slow connection the process may take quite a lot of time thereby triggering suspend.

castrojo:
## Problem

Users reported that during Bluefin installation and initial user account setup (GNOME Initial Setup), the system would suspend if the process took too long. When the system attempted to resume from suspend, it would fail to recover properly, resulting in:
- Empty screen with keyboard input but no display
- No disk activity
- Systemd failures

This was particularly problematic during:
1. The live ISO installation process (Anaconda installer)
2. First boot user account creation (GNOME Initial Setup)

## Solution

This PR disables power management (suspend/sleep) completely during the live environment and initial setup by adding GNOME GSettings schema overrides to the ISO configuration scripts.

### Changes Made

Modified two ISO configuration scripts:
- `iso_files/configure_iso_anaconda.sh` - Standard Bluefin ISO
- `iso_files/configure_lts_iso_anaconda.sh` - Bluefin LTS ISO

### What the Fix Does

Added a new GSettings schema override file (`zz3-bluefin-installer-power.gschema.override`) that configures:

1. **Disable AC suspend**: `sleep-inactive-ac-type='nothing'`
2. **Disable battery suspend**: `sleep-inactive-battery-type='nothing'`
3. **Zero timeout for AC**: `sleep-inactive-ac-timeout=0`
4. **Zero timeout for battery**: `sleep-inactive-battery-timeout=0`
5. **Disable idle screen blank**: `idle-delay=uint32 0`

These settings apply only to the live ISO environment and GNOME Initial Setup. After the user completes initial setup and logs in for the first time, the normal Bluefin power management settings will apply from the installed system image.

## Benefits

- ✅ Prevents installation failures due to suspend issues
- ✅ Improves user experience during setup
- ✅ No manual intervention required
- ✅ Works for both AC and battery power
- ✅ Consistent behavior across standard and LTS versions
- ✅ Minimal, surgical modification (only 13 lines per file)

## Testing Recommendations

1. Build test ISO with these changes
2. Install on a laptop (battery powered)
3. During installation, let the system sit idle for 20+ minutes
4. Verify the system does not suspend
5. Complete installation and user setup
6. After first login, verify normal power management works

Fixes #2590

### DIFF
--- a/iso_files/configure_iso_anaconda.sh
+++ b/iso_files/configure_iso_anaconda.sh
@@ -20,6 +20,19 @@ welcome-dialog-last-shown-version='4294967295'
 favorite-apps = ['anaconda.desktop', 'documentation.desktop', 'discourse.desktop', 'org.mozilla.firefox.desktop', 'org.gnome.Nautilus.desktop']
 EOF
 
+# Disable suspend/sleep during live environment and initial setup
+# This prevents the system from suspending during installation or first-boot user creation
+tee /usr/share/glib-2.0/schemas/zz3-bluefin-installer-power.gschema.override <<EOF
+[org.gnome.settings-daemon.plugins.power]
+sleep-inactive-ac-type='nothing'
+sleep-inactive-battery-type='nothing'
+sleep-inactive-ac-timeout=0
+sleep-inactive-battery-timeout=0
+
+[org.gnome.desktop.session]
+idle-delay=uint32 0
+EOF
+
 # don't autostart gnome-software session service
 rm -f /etc/xdg/autostart/org.gnome.Software.desktop
 
--- a/iso_files/configure_lts_iso_anaconda.sh
+++ b/iso_files/configure_lts_iso_anaconda.sh
@@ -20,6 +20,18 @@ welcome-dialog-last-shown-version='4294967295'
 favorite-apps = ['anaconda.desktop', 'documentation.desktop', 'discourse.desktop', 'org.mozilla.firefox.desktop', 'org.gnome.Nautilus.desktop']
 EOF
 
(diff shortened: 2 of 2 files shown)


hanthor (commented):
I'm wondering if we instead mask the systemd target while anaconda is running, that way the live environment can be .ore useful?

... But who would need their live usb to suspend?

castrojo:
@copilot reinvestigate masking the services and give us recommendations.

tulilirockz:
Force merge as this PR is a ISO pr and we are not building images on these (needs to be fixed)

Issue #3318, opened 2025-10-04
Switch to bazaar flatpak


castrojo:
It has a new glycin dep that is rawhide only in the RPM. LTS is already on the flatpak we should switch over all bluefins. 

It'd be nice to do this before GTS drops.

hanthor:
We need the curated section working with the flatpak.

Did you already figure this out @renner0e

castrojo:
We're waiting on an upstream release of flatpak which will have the /etc/preinstall feature so that we can manage the lifecycle of flatpaks. That will allow us to have a framework for adding and removing flatpaks in the entire image.

renner0e:
> We need the curated section working with the flatpak.
> 
> Did you already figure this out [@renner0e](https://github.com/renner0e)

yup, just look at my custom image
https://github.com/ublue-os/aurora/issues/948#issue-3394721391
https://github.com/renner0e/diffuse

PR #3352, opened 2025-10-10
feat: add the needed components for flatpak bazaar

renner0e:
# Merge this for the beta only!

Related: https://github.com/ublue-os/aurora/issues/948
partially addresses: https://github.com/ublue-os/bluefin/issues/3318

~~has to be merged pretty much at the exact same time as: https://github.com/ublue-os/bluefin/pull/3307
because we need to adjust the paths in the config for bazaar to not blow up.~~

I made it so that we now ship 2 bazaar configs
- one that lives in /etc which the flatpak  will use (and the rpm can't use!), that will have the fancy banners with the dark/lght switching and jxl images
- and the old one used by rpm in /usr, the plan is to remove this config when the transition is over!

Trivial changes like adding new apps would have to be done for both configs (or we just don't care about the rpm one anymore :P)

~~We can technically remove the rpm from the image on this PR if we want to, if this transition fails for whatever reason they still have the RPM that should just work (the config should be misisng tho)~~

On the next flatpak release (1.17?) it will be possible to easily migrate existing rpm bazaar users to the flatpak.

I don't know if you need to do anything on the gnome side of things, for it to show up in the dock or whatever.

The permissions override + tmpfiles symlink thing and service unit for bazaar we already use on bluefin-lts.

Would be nice to get some feedback on the systemd service, in regards to error handling (what if the user has no network, how long should we wait after the boot, how many times is it allowed to fail and so on)

This service unit is not blessed by upstream in any way.

As it currently works, bazaar will not be installed by the time the service wants to kick off (this will vary on the network and so on), this is only an issue on the first boot into the image with this change. So it would be nice if we make the first boot experience of this transition as smooth as possible.

### DIFF
--- a/build_files/base/05-override-install.sh
+++ b/build_files/base/05-override-install.sh
@@ -31,7 +31,7 @@ if [[ "${UBLUE_IMAGE_TAG}" == "beta" ]]; then
   dnf5 -y --repo=copr:copr.fedorainfracloud.org:ublue-os:flatpak-test swap flatpak-libs flatpak-libs
   dnf5 -y --repo=copr:copr.fedorainfracloud.org:ublue-os:flatpak-test swap flatpak-session-helper flatpak-session-helper
   # print information about flatpak package, it should say from our copr
-  rpm -q flatpak --qf "%{NAME} %{VENDOR}\n" | grep -F -e "flatpak-test"
+  rpm -q flatpak --qf "%{NAME} %{VENDOR}\n" | grep ublue-os
 fi
 
 # Offline Bluefin documentation


system_files/shared/usr/share/flatpak/preinstall.d/bazaar.preinstall
p5, line 1: So to clarify, this means Bazaar will be pulled from the internet on first boot?  

Is there a way we can make the offline ISO Flatpaks work well with this? 

We probably need to disable this service inside the live environment too.

castrojo: Yeah we still need the thing offline, but I am assuming we would include it on the ISO and this step would just noop?

renner0e: > Is there a way we can make the offline ISO Flatpaks work well with this?
> 
> We probably need to disable this service inside the live environment too.

Thank you both, I completely forgot about the ISOs, I pushed a commit fixing that, disabled the service on the iso as well

> So to clarify, this means Bazaar will be pulled from the internet on first boot?

Correct, on every boot it will execute `flatpak preinstall`  and check for .preinstall files, if there are flatpaks that have not been installed yet, then they will be installed.

My only "issue" with it currently is the first boot into the first image that ships this change, so on this boot specifically where bazaar is not immediately present (because flatpak preinstall has not finished yet, bazaar is not installed yet and this varies depending on the users machine) the background service obviously can't run. But it's not a super huge deal in my opinion, they can just run it manually and it will work.

So I expect the first boot to be broken in the way that the gnome search thingy will not work because the systemd service timed out because of too many failures.

But on the next boot everything (or whenever the bazaar service runs) should be fine. If users are offline during the first boot then they'll obviously not have bazaar and bazaar (rpm) will be gone from a user POV (doesn't matter anyway as it's useless without a network connection).

It's quite cleverly designed and it will not just blindly install bazaar on every boot or something (over the network!).
So it will just do this in like 99% of boots

```
flatpak preinstall -y
Nothing to do.
```


If you remove bazaar (because you don't like it or whatever) you essentially opt-out of this preinstall thing (for bazaar). There is a config file that stores which flatpaks already have been preinstalled, it will only do the whole preinstall thing once.

So you will not get into a situation where you remove bazaar and then on the next boot you are getting forced into installing it again.

This also means if we remove the bazaar preinstall file bazaar will get uninstalled on the users system, unless the user has explicitly installed it.

To quote the [manpage](https://github.com/flatpak/flatpak/pull/6116/files#diff-e91a0e3ed530b6ebc97d05c0a5e7b19fced3cdf6ec3a87ee9c9f50a04f82353b):

```
Preinstalled flatpaks are defined by dropping .preinstall files into the directories
/usr/share/flatpak/preinstall.d/ and /etc/flatpak/preinstall.d/. The OS runs flatpak
preinstall -y (or its GUI equivalent) on system startup, which then does the actual
installation.
``` 

We have to take care of the OS part by having a systemd service or similar, there is not a service that just gets shipped with the new version or something that takes care of that automatically.

I hope my ramblings explains this situation quite  well. It's 3 AM so excuse me lol

Thank you for coming to my TED talk.

renner0e:
more details discussed here:
https://github.com/ublue-os/aurora/pull/1076#issuecomment-3395206131

castrojo:
Couldn't we just do a flatpak-beta.sh and then do an F43-only if in there and then add the copr?

renner0e:
> Couldn't we just do a flatpak-beta.sh and then do an F43-only if in there and then add the copr?

sure, should I do that?

renner0e:
so my last commit [`7ee8765` (#3352)](https://github.com/ublue-os/bluefin/pull/3352/commits/7ee8765d8fa7b0bf78d9013a14cd17c0ded50ef8) still leaves the systemd tmpfiles stuff for the bazaar permission on the stable image, this should be fine and not break anything.

build_files/base/05-override-install.sh
```diff
@@ -25,6 +25,16 @@ if [[ "${FEDORA_MAJOR_VERSION}" -lt "43" ]]; then
         fwupd fwupd
 fi
 
+# TODO: remove me on next flatpak release when preinstall landed
+if [[ "${UBLUE_IMAGE_TAG}" == "beta" ]]; then
+  dnf5 -y --repo=copr:copr.fedorainfracloud.org:ublue-os:flatpak-test swap flatpak flatpak
+  dnf5 -y --repo=copr:copr.fedorainfracloud.org:ublue-os:flatpak-test swap flatpak-libs flatpak-libs
+  dnf5 -y --repo=copr:copr.fedorainfracloud.org:ublue-os:flatpak-test swap flatpak-session-helper flatpak-session-helper
+  # print information about flatpak package, it should say from our copr
+  rpm -q flatpak --qf "%{NAME} %{VENDOR}\n"
```
tulilirockz:
```suggestion
  rpm -q flatpak --qf "%{NAME} %{VENDOR}\n" | grep -F -e "flatpak-test"
```

castrojo (approved):

PR #3360, opened 2025-10-12
feat: add ews support

jumpyvi:
# Add support for EWS

## What

Adds the evolution-ews-core package, this does NOT install the full evolution suite

## Why

- Even if we prefer FOSS solutions, many organisations and individuals use Microsoft 365, having integration with Microsoft calendar and mail is important for many. From my testing it works in the Calendar flatpak and in the top bar calendar dropdown, it does not seem to work with endeavour.
- The option for calendar/mail is there in gnome-online-account, you can activate it, but nothing will happen if you don't have ews installed. This is really confusing behaviour if you don't know how Gnome manages online accounts.

*Thank you JF for teaching me about the existing of the  -core package*

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -16,6 +16,7 @@
 				"cryfs",
 				"davfs2",
 				"ddcutil",
+				"evolution-ews-core",
 				"evtest",
 				"fastfetch",
 				"firewall-config",


jumpyvi:
It either need be exluded from 41 or included only in 42, which way should it be done?

castrojo:
Ok we got this in while doing Beta/F43 enablement, thanks for the help!

hanthor:
Do we need to add evolution-ews as a flatpak addon?

Issue #3353, opened 2025-10-10
Turtle - Nautilus git manager (for developer aka -dx images)


rugk:
### Describe the package

I'd like to request the paI'd like to request the package Turtle in the dev images because I guess having a git integration in your file manager is likely a thing maaaany developers want.

![screenshot of integration](https://gitlab.gnome.org/philippun1/turtle/-/raw/main/img/emblems.png)

At least, I want to spark discussion about it.

See https://gitlab.gnome.org/philippun1/turtle

Note it is available on flathub as a flatpak: https://flathub.org/apps/de.philippun1.turtle

**But it needs additional components installed** due to the fact it needs to integrate into system nautilus, so I am not sure whether a flatpak installation would be most preferred in this case.
As [the installation button/guide suggests](https://gitlab.gnome.org/philippun1/turtle#flatpak-plugin-requirements) most system packages were already installed for me, only `meld` was missing (as a dependency).
However, you also need to install the actual plugin (into `~/.local/share/nautilus-python/extensions` AFAIK)
This may be too much to properly and safely integrate/manage and as you likely want to have it fully automated maybe a system package would be better?

But yeah, I am open to discussion, just want to let you know of this great tool!

---

Accidentally opened in wrong distro repo: https://github.com/ublue-os/aurora/issues/1069

mheuvel-dev:
Thanks for bringing up this nice plugin.

I just did the flatpak install. 

After installing the flatpak, open the `Turtle Settings` app and press `setup` next to the `Install plugin`.

<img width="795" height="717" alt="Image" src="https://github.com/user-attachments/assets/fd3473fb-5a59-4713-8dba-5b4846e47c1d" />

It works like a charm:

<img width="865" height="692" alt="Image" src="https://github.com/user-attachments/assets/c5539ec8-0333-4f74-bc38-7c6394d8d27f" />

Edit: you're right, I did have `meld` installed as flatpak already and that works together with Turtle.

castrojo:
This would be good to add to the developer section to bazaar if someone wants to do that. Feel free to send a PR!

PR #3365, opened 2025-10-12
feat: Add Turtle to Developer section in Bazaar

ExistingPerson08:
This PR adds [Turtle](https://flathub.org/en/apps/de.philippun1.turtle) to developer section in Bazaar, as mentioned in https://github.com/ublue-os/bluefin/issues/3353.


mheuvel-dev:
Nice!

I'm not sure if this is already possible and how much effort it is to trigger some post-install script in bazaar.
Users should now go to the GUI to finalize the `setup`, or we could trigger a post install as a service:

```
flatpak run de.philippun1.turtle plugin_installer 
```

In addition, Turtle uses `https://flathub.org/en/apps/org.gnome.meld` as a diff and merge tool.
Do we want to somehow bundle the install, or do we leave this to the user itself?

Leaving as is is also fine - I'm happy to help if you have some hints on how to do this.

castrojo:
Let's just add them both, easy win, and then we can investigate better integration later. Thanks!

PR #3404, opened 2025-10-17
feat: enable bazaar flatpak on beta tags

ahmedadan:
Introduces beta-specific flatpak lists that are installed in addition
to standard flatpaks when building with UBLUE_IMAGE_TAG=beta. Adds
Bazaar to beta channel only.

### DIFF
--- a/build_files/shared/build-dx.sh
+++ b/build_files/shared/build-dx.sh
@@ -12,13 +12,6 @@ cp /ctx/packages.json /tmp/packages.json
 # Copy DX flatpak list
 install -Dm0644 -t /etc/ublue-os/ /ctx/flatpaks/system-flatpaks-dx.list
 
-# Copy DX beta flatpaks if this is a beta build
-if [[ "${UBLUE_IMAGE_TAG}" == "beta" ]]; then
-    if [[ -f /ctx/flatpaks/system-flatpaks-dx-beta.list ]]; then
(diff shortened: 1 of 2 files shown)


renner0e:
So can I remove my commits related to adding them to the flatpak list on my branch?

renner0e (commented):
I don't understand, why is there a need for a DX-specific list?
We don't make dx isos anymore.
The DX flatpak list is used by the devmode just recipe IIRC and bazaar is not developer specific.

ahmedadan:
> I don't understand, why is there a need for a DX-specific list? We don't make dx isos anymore. The DX flatpak list is used by the devmode just recipe IIRC and bazaar is not developer specific.

Ah good to know, I just made an assumption last night about the need for both lists. I'll make a commit to fix this shortly.

ahmedadan:
> So can I remove my commits related to adding them to the flatpak list on my branch?

I was unaware this was already being done, @castrojo asked for some support in Discord and I made a branch.

When do you think your branch will land? If there's no value in this PR I can close it.

castrojo:
The issue is renner figured out what to do but we don't know how to land this in beta? I thought we needed both?

castrojo:
Ok I fear I have sent you on a wild raptor chase. The original PR is merging now after we made some adjustments, thanks for helping out!

PR #3443, opened 2025-10-21
chore: add step to install Brewfile dependencies

hanthor:
Added a step to install dependencies from Brewfiles in the workflow.


castrojo (approved):
Thank you I was struggling with this locally.

PR #3446, opened 2025-10-21
fix: remove terra repos

castrojo:
Let's remove this, this is upstream now in GNOME, we can tell people to move to stable if they really want this, better than keeping another repo around.

### DIFF
--- a/build_files/base/04-packages.sh
+++ b/build_files/base/04-packages.sh
@@ -43,8 +43,8 @@ FEDORA_PACKAGES=(
     gum
     hplip
     ibus-mozc
-    igt-gpu-tools
     ifuse
+    igt-gpu-tools
     input-remapper
     iwd
     jetbrains-mono-fonts-all
@@ -78,6 +78,7 @@ FEDORA_PACKAGES=(
     sssd-ad
     sssd-krb5
     sssd-nfs-idmap
+    switcheroo-control
     tailscale
     tmux
     usbip


build_files/base/04-packages.sh
```diff
@@ -194,28 +194,6 @@ if [[ "${#EXCLUDED_PACKAGES[@]}" -gt 0 ]]; then
...
-
-# shellcheck disable=SC2016
-if [[ "${FEDORA_MAJOR_VERSION}" -lt "43" ]]; then
-    dnf -y swap \
-        --repo=terra --repo=terra-extras \
-        gnome-shell gnome-shell
-    dnf versionlock add gnome-shell
-    dnf -y swap \
-        --repo=terra --repo=terra-extras \
-        switcheroo-control switcheroo-control
```
tulilirockz, line 215: I think we do still need the switcheroo-control package, the server-side part of it is fine though!

tulilirockz: We can get that package via the fedora repos though i believe

tulilirockz:
Re added switcheroo-control, but straight from the repos. Also sorted the packages array

castrojo:
Should we merge this since switcheroo-control has been in distro already? It's also in F41, I didn't notice that.

inffy:
> Should we merge this since switcheroo-control has been in distro already? It's also in F41, I didn't notice that.

I think you will still need the patched version for F42, if you want that version that is. Seems fedora won't push the 3.0 version (which has the patch) to F42, only F43.

PR #3448, opened 2025-10-21
chore: delete nvidia-closed support

tulilirockz:
We've been talking about this for a while now, this is a DRAFT! Please only merge when completely sure

This removes all the handling for nvidia-closed specific images, and stops building the images, this does NOT(!) move anyone to nvidia-open, as this would be more of a breaking operation for people on older hardware. I believe the best approach is having them not get system updates than not being able to use their system at all because of nvidia "breaking"


dylanmtaylor:
This is an awesome change, but I think there should be a migration post encouraging a rebase for existing users too. This is kind of hidden right now.

castrojo:
Yeah we'll mention it again in the release notes, the deprecation was announced last cyle. Don't have a link handy but most of the package pulls as ci noise, it's not that many. The ISOs for these were removed 6 months ago, etc. etc. The 30 day notice will fire off so we can handle the oneseys and twosies as they come in.

PR #3451, opened 2025-10-22
feat(just):: Add install-fonts command to justfile

castrojo:


castrojo:
This is ready. I don't think adding a check like copilot says matters since both would result in an error anyway. I've updated the Bluefin docs to mention these commands. Look how awesome the apps list looks like now:

https://docs.projectbluefin.io/command-line#bluefin-cli

tulilirockz (approved):

PR #3455, opened 2025-10-22
fix(beta): remove bazaar rpm

castrojo:
Comment out bazaar installation for case 43 in the script.


renner0e (changes requested):
*If you also want to enable this for the stable images*:

I'd be more in favor of doing it the safe way with the transitionary period for a couple weeks but it has worked on the beta so yolo it I guess.

No one seems to have complained about flatpak not working so far. So we should be good in this backported patch regard.

Could you add the changes from https://github.com/ublue-os/bluefin/pull/3456 to have that in one swoop?

- we can also remove the libdex override
- enable the service in 17-cleanup.sh (uncomment)
- remove /usr/share/ublue-os/bazaar, we don't need it

merge on sunday after the weeklies

I think that should be all

renner0e (commented):
https://github.com/ublue-os/bluefin/pull/3455/files#diff-be4b9c415b1fed80d5c342409f5551c28c976d983dbaa9609ba09abee88f10e5R29-R30
I guess you can remove this now

renner0e (commented):
this has to run on F42+ images then
https://github.com/ublue-os/bluefin/pull/3455/files#diff-be4b9c415b1fed80d5c342409f5551c28c976d983dbaa9609ba09abee88f10e5L31-L34

renner0e (commented):
we need to run the dnf swap for our backported flatpak for F42+
if we don't then users will not have bazaar rpm and it won't be automatically installed

PR #3456, opened 2025-10-22
feat: add bazaar.service for background service management

castrojo:


renner0e:
I tested this and this is still needed btw, there is no dbus activation on the bazaar side or anything

PR #3480, opened 2025-10-25
feat: hack webui to overwite title

hanthor:


hanthor:
Doesn't work anyway. Will ned to figure out something else

castrojo:
Ok let's revisit later.

PR #3504, opened 2025-10-28
container: fix dracut breaking because of different filesystems under /var/tmp and /tmp

tulilirockz:
I dont wannaaa.....


tulilirockz:
Force merged as GTS builds are breaking because ublue COPR disabled f41

PR #3512, opened 2025-10-28
cleanup: remove/mask fedora flatpaks service directly on bluefin as well

tulilirockz:


build_files/shared/clean-stage.sh
```diff
@@ -6,6 +6,9 @@ set -eoux pipefail
 
 dnf clean all
 
+systemctl mask flatpak-add-fedora-repos.service
+rm -rf /usr/lib/systemd/system/flatpak-add-fedora-repos.service
```
alatiera:
```suggestion
rm -f /usr/lib/systemd/system/flatpak-add-fedora-repos.service
```

tulilirockz: True, good idea

PR #3517, opened 2025-10-29
preinstall: Add CollectionId to the config

alatiera:
Add a CollectionID so flatpak will only look for the application on Flathub.


alatiera:
This is Draft cause Flathub doesn't actually publish its CollectionId yet, and currently needs to be manually done.

`flatpak remote-modify --collection-id=org.flathub.Stable flathub`

castrojo (approved):
Adds `CollectionID=org.flathub.Stable` to the Bazaar preinstall config so flatpak only resolves the app from Flathub. One-line correct fix. LGTM.

castrojo:

PR #3518, opened 2025-10-29
feat: automate ISO builds monthly for GTS, LTS, and stable

castrojo:
DO NOT MERGE until after more ISO testing. 

Schedule Live Anaconda ISO workflow to run on the first day of each month at 2:00 AM UTC. Automatically builds GTS, LTS, and stable ISOs for both main and nvidia-open flavors (where applicable), plus LTS gdx variants for both amd64 and arm64 platforms.


castrojo:
This will be replaced by @projectbluefin/iso

PR #3521, opened 2025-10-29
fix: nuke fedora flatpak repo from orbit

alatiera:
A stray update introduced flatpak-add-fedora-repos.service and now the repositories are setup in a lot of existing installation. Nuke them completly on every boot just to be sure.

fixes: https://github.com/ublue-os/bluefin/issues/3520


system_files/shared/usr/lib/systemd/system/flatpak-nuke-fedora.service
```diff
@@ -0,0 +1,20 @@
...
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/flatpak remote-delete --system fedora 
+ExecStart=/usr/bin/flatpak remote-delete --system fedora-testing  
+# Flatpak will fail if the remote doesn't exit, but we don't mind
+SuccessExitStatus=1
+# Make sure even if flatpak-add-fedora-repos.service  exists, it
+# won't run.
+ExecStartPost=/usr/bin/touch /var/lib/flatpak/.fedora-initialized
```
alatiera: This should be just ExectStart as well

PR #3564, opened 2025-11-01
feat(flatpaks): replace Clapper with Showtime

LorbusChris:
Clapper is not well maintained and still pulls in the GNOME 47 flatpak runtime.
See https://github.com/flathub/com.github.rafostar.Clapper/issues/61

Let's switch to GNOME Showtime which is well maintained and the official video player for this DE.

Docs PR: https://github.com/ublue-os/bluefin-docs/pull/459


castrojo:
Can you update bluefin-docs too? should be a simple search and replace, thanks!

PR #3568, opened 2025-11-01
feat(flatpaks): add more apps

LorbusChris:
Add more bespoke apps to the curated bluefin lists.

First commits sorts the lists.
One commit for each app addition.

I deem all of these apps uniquely useful. For each I'd be happy to provide an even better justification, or drop it. Maintainers, please lets discuss.

I shall update the docs with sponsor links when this merges.

### DIFF
--- a/flatpaks/system-flatpaks-dx.list
+++ b/flatpaks/system-flatpaks-dx.list
@@ -6,4 +6,3 @@ app/me.iepure.devtoolbox
 app/org.freedesktop.Bustle
 app/org.gnome.dspy
 app/org.pipewire.Helvum
-app/sh.loft.devpod
--- a/system_files/shared/etc/bazaar/config.yaml
+++ b/system_files/shared/etc/bazaar/config.yaml
@@ -330,7 +330,6 @@ sections:
       - com.mardojai.ForgeSparks
       - io.dbeaver.DBeaverCommunity
       - cc.arduino.IDE2
-      - sh.loft.devpod
       - dev.geopjr.Calligraphy
       - io.github.lo2dev.Echo
       - org.gnome.design.IconLibrary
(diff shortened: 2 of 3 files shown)


flatpaks/system-flatpaks.list
```diff
@@ -31,6 +31,7 @@ app/org.gnome.Showtime
 app/org.gnome.SimpleScan
 app/org.gnome.TextEditor
 app/org.gnome.Weather
+app/org.gnome.World.Secrets
```
LorbusChris: maybe GNOME Secrets should be moved to extras

castrojo:
Adding to the system flatpaks means they get added to the ISO, so we keep that frozen. If you can put everything in the extras files that will be fine. dx changes look good.

The extras section will eventually be a dedicated section in Bazaar with one "install all of these" button so we want all the apps in here. Though we should consider moving things out of the system flatpak list and into extras. That would take a community discussion, open an issue on that one we should perhaps take a closer look. 

We're for sure not adding extensions though so please drop that.

LorbusChris:
Updated. I also added a commit that removes DevPod, since that's unmaintained and it pulls in ROL GNOME 47 platform stuff

castrojo:
These look great, a few are too niche:

- Let's not ship dconf-editor - if you have to use that tool you're already miserable let's not ship that.
- turnon is too niche, this would be cooler in the curated section IMO
- Feels like Snapshot should actually be a default app now that I look closer at it.
- I'm not seeing the use case for seahorse, we don't care about gpg and ssh keys are a cli thing. Are people using this? 

Everything else lgtm, thanks!

LorbusChris:
re: dconf-editor, OK, I'll drop it

re: Turnon, can't argue that it's not niche lol, will drop it too. 

re: Snapshot, will move back to system-flatpaks

re: Seahorse.  It's not just GPG and SSH keys. Seahorse can manage the entire GNOME Keyring, i.e. app, wifi, samba and other secrets and passwords stored there. It's been a thing in GNOME forever. Please give this another thought. If you'd still like me to remove it, I'll do that.

castrojo:
I'm just confused on the use case for seahorse. I don't use seahorse and all my secrets are managed fine. But fine whatever leave it in there. 😄

LorbusChris:
I kinda use it like the passwords app on iOS. From time to time I need to lookup a password from some app, and Seahorse makes that super easy.

castrojo (approved):

castrojo:
None of the failures are because of this so merging.

Issue #3569, opened 2025-11-02
Add renovate rules to update gnome extensions


castrojo:
The renovate rules need to account for the git submodules for gnome extensions. It should check for the latest stable release of each extension.

castrojo:
The blur my shell update isn't being picked up. Also does anyone have caffeine working in f43? that should be working.

bdflemin:
@castrojo I believe that the submodule entry here is working as expected in F43 as its working for me now and using the correct module version:

https://github.com/ublue-os/bluefin/blob/main/.gitmodules#L9-L12

<img width="998" height="481" alt="Image" src="https://github.com/user-attachments/assets/d1b5fe9a-095f-40a4-a7d5-5f5201af23d1" />

The only one not working is the `blur my shell` extension which got a PR update from the maintainer but something to keep an eye on:

Issue merged to help with GNOME 49 - https://github.com/aunetx/blur-my-shell/pull/791
v70 should help with GNOME 49 - https://github.com/aunetx/blur-my-shell/releases/tag/v70

castrojo:
Yeah we're on 69 of blur but our renovate rule isn't picking up that upgrade

https://github.com/ublue-os/bluefin/blob/82bb66e6232594376fec6b268ddbc92a07dd9c68/.gitmodules#L5

bdflemin:
I can confirm on the latest stable image today that all the GNOME extensions seem to be looking 'better' for the ones that seem to be having issues. For this issue I see we have the following in place to build out and install for future builds:

* AppIndicator/KStatusNotifierItem
* blur-my-shell
* caffeine
* dash-to-dock
* gsconnect
* logo menu
* search light

Not sure if we want to add any more to this but I believe we are fine for now until we know more about `blur-my-shell` and where that extension goes for the future if the maintainer doesn't want to maintain anymore 😢

PR #3598, opened 2025-11-04
updating to v70 on the blue-my-shell extension for GNOME 49 support

bdflemin:
Yesterday the gnome extension called `blur-my-shell` was updated to allow for gnome 49 support and helps with #3569  issue as well to keep it the extension alive for now.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,7 +5,7 @@
 [submodule "system_files/shared/usr/share/gnome-shell/extensions/blur-my-shell@aunetx"]
 	path = system_files/shared/usr/share/gnome-shell/extensions/blur-my-shell@aunetx
 	url = https://github.com/aunetx/blur-my-shell.git
-	branch = v70
+	branch = master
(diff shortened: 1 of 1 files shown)


.gitmodules
p5, line 1: Please can you update the submodule commit too.  The `system_files/shared/usr/share/gnome-shell/extensions/blur-my-shell@aunetx` still references the old branch/repo

I _think_ it needs

```
git submodule sync --recursive
```

bdflemin: So I found that the project is not using branches but tags for the releases, I will need to make some changes to make this work. Right now its on `master` which I will be changing to but also will see if I can make sure its on tag v70 later tonight

bdflemin: ok was able to make sure the submodule will use master branch and it would delegate tag v70 to make sure gnome 49 will work.

Sorry for the delay on this, had to help with oldest's baseball practice.

castrojo (approved):

PR #3717, opened 2025-11-26
fix: update bug-report.yml

castrojo:
Add some info so people can be told up front to share their image info so we can confirm issues faster.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -39,4 +39,4 @@ body:
     id: extra-context
     attributes:
       label: Extra information or context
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -39,4 +39,4 @@ body:
     id: extra-context
     attributes:
       label: Extra information or context
-      description: Add any other context about the problem here.
+      description: IF YOU ARE REPORTING A REGRESSION: Use `ujust rebase-helper` to find the version that used to work. Report the working version and the broken version so that other users can rebase and confirm the issue!
```
tulilirockz:
```suggestion
      description: If you are reporting a regression: Use `ujust rebase-helper` to find the version that used to work. Report the working version and the broken version so that other users can rebase and confirm the issue!
```
This sounds oddly aggressive, should we do that?

PR #3736, opened 2025-11-28
fix(uupd): apply disable-module-distrobox on all versions

befanyt:
Currently distrobox updates are only disabled on F42. This was introduced in https://github.com/ublue-os/bluefin/pull/2458 and gated because of the availabilty of uupd. 

From now on all images have uupd, so the gate could be removed.


tulilirockz (approved):

castrojo:
@befanyt your contributions have always been excellent and regression-free. I appreciate the pride you put into your work, interested in becoming a comaintainer?

PR #3751, opened 2025-12-01
chore: clean up a bunch of stuff from the image

tulilirockz:
This is basically an equivalent of [what was done on ublue-os/aurora](https://github.com/ublue-os/aurora/pull/1332) a few days ago, we should save a few hundred megabytes by removing these things from the image


tulilirockz:
Drafted because I do not know the scope of what may break yet

renner0e:
no one complained about this so far on Aurora so bluefin can definitely get away with this

Issue #3710, opened 2025-11-24
Yubikey (GPG key support) broken in Fedora Silverblue 42. Should this be fixed in Bluefin?


castrojo:
### Discussed in https://github.com/ublue-os/bluefin/discussions/3681

<div type='discussions-op-text'>

<sup>Originally posted by **NiklasMerz** November 20, 2025</sup>
Hey there,

I just struggled to get my Yubikey working again and came across this thread: https://discussion.fedoraproject.org/t/gpg-with-yubikeys-on-fedora-43-scdaemon-removed/170981

Installing `gnupg2-scdaemon` with rpm-ostree. It's the first time I needed to use rpm-ostree since everything else I need is wonderfully handled by Bluefin.

Maybe it's just me but I think GPG smartcard support could  be important for the Bluefin developer audience so I'd like ask if better support by installing this in Bluefin is a good idea.

Quote from the forum. Maybe Bluefin can offer a good out of the box experience:

> As an aside: getting Yubikeys working is always a fuck-around, on all OSs. It would be cool if they worked out-of-the-box in Fedora.

</div>

klmcwhirter:
> @castrojo Maybe it's just me but I think GPG smartcard support could be important for the Bluefin developer audience so I'd like ask if better support by installing this in Bluefin is a good idea.

FWIW, I tend to agree with you even though I have not started using a smartcard yet. The main reason is there are frustratingly few services I use that support them. Banks, Insurance Companies, Utility companies, for example still seem to be in the Model A days. I feel like I have to _turn the crank_ to log in. [i.e., receive OTP via SMS]

repires:
discussion: https://github.com/ublue-os/bluefin/discussions/3681

Fixed by: https://github.com/ublue-os/bluefin/pull/3774

PR #3774, opened 2025-12-05
fix: add gnupg2-scdaemon

castrojo:


renner0e:
Should fix: https://github.com/ublue-os/bluefin/issues/3710

renner0e (approved):

PR #3815, opened 2025-12-11
fix: remove kernel-devel from rpmdb since /usr/src is removed from image

salaniz:
## Description
This commit (https://github.com/ublue-os/bluefin/commit/728e775c8b806755ef8c41f56c71caa5363f222c) removed the `/usr/src` directory from the image. However, `kernel-devel` among other packages that created files in this directory are still marked as installed in the image. This causes issues when bluefin is used as a base image for installing additional packages (as per https://github.com/ublue-os/image-template) that depend on `kernel-devel` as the package installation will fail with
```
Could not find files needed to compile modules for 6.17.7-200.fc42.x86_64
Are the development files for kernel 6.17.7-200.fc42.x86_64 or the appropriate kernel-devel package installed?
ERROR: kernel or kernel-devel required for 6.17.7-200.fc42.x86_64
```

As a solution, this PR removes `kernel-devel` from the rpmdb such that it can be reinstalled in custom images if it is required as a dependency.

Note that the added lines conceptually fit better into `build_files/base/04-packages.sh`, but I decided to add it directly after the files are removed in `build_files/base/05-override-install.sh` to keep it logically connected and easier to maintain.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Build/CI improvement

## Testing Done
- Local build: No
- Tested on running system: No
- Just recipes tested: No

## Related Issues
None

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -44,6 +44,18 @@ jobs:
         with:
           submodules: recursive
 
+      # We use both actions as stable nvidia-open-dx failed due to /var/tmp being too small
+
+      # mount /mnt as /var/lib/containers
+      - name: Mount BTRFS for podman storage
+        id: container-storage-action
+        uses: ublue-os/container-storage-action@911baca08baf30c8654933e9e9723cb399892140
+        continue-on-error: true
+        with:
+          target-dir: /var/lib/containers
+          mount-opts: compress-force=zstd:2
+          loopback-free: '1'
+
       - name: Maximize build space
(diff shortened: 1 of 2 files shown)


castrojo (approved):

salaniz:
It might be safer to only remove the kernel files `rm -rf /usr/src/kernels` so not to run into unforeseen issues in case a future package installs something into `/usr/src`.
Currently, the only other directory seems to be an empty `/usr/src/debug` directory so the savings should remain.

At the same time, I can see that there might be limited use cases for having source files in the image (custom images, rpm-ostree installs) and it currently guarantees that everything is cleaned.

PR #3846, opened 2025-12-14
chore: move `ublue-privileged-setup` references to `/usr/bin` instead of `/usr/libexec`

tulilirockz:
Required for migration to https://github.com/projectbluefin/common.git -
This is closer to the FHS standard instead of being fedora-specific

Needs to be merged once https://github.com/projectbluefin/common/pull/49 is merged as well


castrojo:
Do we still need this?

castrojo (approved):

PR #3857, opened 2025-12-16
chore: move system files to common

renner0e:
these are all the files that we in my opinion can move to common, so deleting them here

More importantly are the files we do not move


renner0e:
https://github.com/ublue-os/bluefin/pull/3959

PR #3888, opened 2025-12-22
fix: remove old motd tips

dancasey:
## Description

Simply remove three motd tips that do not work anymore. These `just` commands have been replaced by corresponding `ujust bbrew` entries. 

- `ujust bluefin-k8s`
- `ujust bluefin-ai`
- `ujust bluefin-fonts`

Perhaps we should make a tip for `ujust bbrew`. Either way, we should stop telling users to try a command that has been removed. 

## Testing Done

None: this is a trivial, docs-type change. 

## Related Issues

None: I didn't see an issue for this yet and have not created one.


castrojo:
Hi Dan, I appreciate the PR, but we're blowing all of these files away soon as we've moved the config over to: https://github.com/projectbluefin/common/blob/main/system_files/bluefin/usr/share/ublue-os/motd/tips/10-tips.md

We should be done soon and I'll update the contributing guide to reflect the new layout, sorry for the confusion, thanks anyway!

PR #3890, opened 2025-12-22
feat: dynamic wallpaper with hemisphere detection

### LINKED ISSUE #1594: default wallpaper's season only work for users from the northen hemisphere
### Describe the bug

the default wallpaper's season is set to the opposite season

the wallpaper's season should be also able to match the season in the southern hemisphere

_No response_

### Output of `groups`

_No response_

_No response_

joshyorko:
## Summary

Adds automatic wallpaper switching based on hemisphere location detection using GeoClue. The wallpaper changes based on whether you're in the northern or southern hemisphere.

**This PR continues the work from #2661** - all credit goes to:
- **@dae3** - Original implementation
- **@m2Giles** - Hook script improvements

The original PR was approved (lgtm) but stalled waiting on feedback. This PR addresses @castrojo's feedback by removing the unnecessary `.desktop` file since the feature runs automatically via systemd timer.

## Changes

- Adds `bluefin-dynamic-wallpaper` script that detects hemisphere via GeoClue
- Adds `get-geoclue-latitude.py` helper for location detection  
- Adds systemd service and timer (runs daily + 1 min after boot)
- Adds user setup hook to enable the service
- **Removes `.desktop` file** - not needed since feature is automated

## Test plan

- [ x] Verify systemd timer activates after boot
- [ x] Verify wallpaper changes based on detected hemisphere
- [ x] Verify graceful handling when location services unavailable

Closes #1594
Supersedes #2661


maphew:
Thanks for this work, I'm looking forward to it. Any chance of being able to optionally fine tune location? There's a big difference in light and dark time across latitudes, especially in the far north and south.

joshyorko:
Thanks for the feedback @maphew! You raise a great point about latitude-based variation in daylight hours.

The current implementation focuses on hemisphere detection to offset the seasonal wallpaper by 6 months (so winter in the Southern Hemisphere shows summer wallpapers and vice versa). This provides the basic seasonal alignment without adding too much complexity.

For latitude-sensitive adjustments (like accounting for polar day/night cycles or twilight duration at high latitudes), we'd need to implement actual sunrise/sunset calculations based on latitude + date, which is a more significant undertaking.

I think that would make an excellent follow-up feature! We could potentially:
- Add optional latitude thresholds for extended twilight adjustments
- Use a library like `suntime` or similar to calculate actual daylight hours
- Allow user overrides via a config file

Would you be interested in opening a feature request issue for this enhancement? That way we can track it separately and keep this PR focused on the core hemisphere detection feature. Happy to collaborate on that follow-up!

castrojo (approved):
Dynamic wallpaper with hemisphere detection via GeoClue. Systemd timer runs daily + 1min after boot. Gracefully falls back to northern hemisphere if location unavailable. Removes the old hardcoded month-based wallpaper logic. Supersedes #2661. LGTM.

castrojo:
Let's fix this in common.

PR #3959, opened 2026-01-02
chore: clean up every file referenced by projectbluefin/common already

tulilirockz:
Sister PR: https://github.com/projectbluefin/common/pull/101

Note that we've also removed quite a few unnecesary things from our repos as well on common directly. All the maybe-breaking changes are separate commits under the sister PR.

### DIFF
--- a/image-versions.yml
+++ b/image-versions.yml
@@ -2,7 +2,7 @@ images:
   - name: silverblue-main
     image: ghcr.io/ublue-os/silverblue-main
     tag: latest
-    digest: sha256:e1f9cd5db8c9b96750db562356ebee3f1d505f568add64d40fa91722191f7c7b
+    digest: sha256:b836644a7f408258b61ca980c40c4da54bc7a735288456ad5a5018188959bbed
   - name: common
     image: ghcr.io/projectbluefin/common
     tag: latest


build_files/shared/build.sh
```diff
@@ -8,7 +8,7 @@ echo "::group:: Copy Files"
 install -Dm0644 -t /etc/ublue-os/ /ctx/flatpaks/*.list
 
 # We need to remove this package here because lots of files we add from `projectbluefin/common` override the rpm files and they also go away when you do `dnf remove`
-dnf remove -y ublue-os-just ublue-os-signing
+dnf remove -y ublue-os-luks ublue-os-udev-rules ublue-os-signing ublue-os-update-services
```
renner0e:
```suggestion
dnf remove -y ublue-os-luks ublue-os-just ublue-os-udev-rules ublue-os-signing ublue-os-update-services
```

we want to keep deleting this

hanthor:
Can we do a diff of /usr and /etc between this post-cleanup Bluefin and the production one?

Then we can see if we are dropping anything important in the final image

tulilirockz:
@hanthor We are explicitly removing a bunch of stuff on this one, still doing it though, sec.

tulilirockz:
We are testing live!

renner0e (commented):
we can also remove the disablement of the services in https://github.com/ublue-os/bluefin/blob/1e2c836ee21e193baf4bb3bfe85b82eec217ffd2/build_files/base/17-cleanup.sh#L35-L37

because we are removing ublue-os-update-services

like here:
https://github.com/ublue-os/aurora/pull/1571/changes

renner0e (dismissed):
LESS GOO

Issue #4048, opened 2026-01-13
Cannot layer packages


lkdm:
### Describe the bug

I can no longer layer packages at all, due to an error

```
~ ❯ sudo rpm-ostree install   1password   1password-cli   ghostty   gtk4-layer-shell   paretosecurity
Checking out tree c4eccbf... done
Enabled rpm-md repositories: updates fedora fedora-multimedia 1password copr:copr.fedorainfracloud.org:scottames:ghostty paretosecurity updates-archive
Importing rpm-md... done
rpm-md repo 'updates' (cached); generated: 2026-01-13T01:06:00Z solvables: 29475
rpm-md repo 'fedora' (cached); generated: 2025-04-09T11:06:59Z solvables: 76879
rpm-md repo 'fedora-multimedia' (cached); generated: 2026-01-13T15:26:28Z solvables: 440
rpm-md repo '1password' (cached); generated: 2025-12-09T11:59:14Z solvables: 2
rpm-md repo 'copr:copr.fedorainfracloud.org:scottames:ghostty' (cached); generated: 2025-11-07T05:45:15Z solvables: 10
rpm-md repo 'paretosecurity' (cached); generated: 2025-09-29T07:14:05Z solvables: 2
rpm-md repo 'updates-archive' (cached); generated: 2026-01-13T01:55:45Z solvables: 64958
Resolving dependencies... done
Checking out packages... done
error: Checkout gnome-shell-extension-appindicator-60-1.fc42.noarch: Hardlinking 01/f17b14a34fc9505e77d5ff1d0ee6666b4321406e34daddcdc113bdace15de4.file to metadata.json: File exists
```

### What did you expect to happen?

I expected the packages to be layered without an error.

### Output of `bootc status`

```shell
● Booted image: ghcr.io/ublue-os/bluefin-dx:gts
        Digest: sha256:735b000c230d532e439f55b5593321f8171154e49a96bc8fb763caec1958ecb6 (amd64)
       Version: gts-42.20260113 (2026-01-13T01:44:56Z)

  Rollback ostree
           Commit: 87c6a39533e83f698f8ee7043d7e70eb103b50ae704226259a538bccb2e5df80
```

### Output of `groups`

```shell
luke wheel docker incus-admin libvirt
```

repires:
Related issue: https://github.com/ublue-os/bluefin/issues/3999

Issue #3999, opened 2026-01-06
Error about fedora-gdm-logo.png during ujust update


woutercoppens:
### Describe the bug

`❯ ujust update
===> Running recipe `update`...
LockLayering=false
note: automatic updates (stage) are enabled
==== AUTHENTICATING FOR org.projectatomic.rpmostree1.upgrade ====
Authentication is required to update software
Authenticating as: wouter
Password:
==== AUTHENTICATION COMPLETE ====
Pulling manifest: ostree-image-signed:docker://ghcr.io/ublue-os/bluefin-dx-nvidia-open:stable-daily
Checking out tree 510b4cf... done
Enabled rpm-md repositories: updates fedora fedora-multimedia updates-archive
Importing rpm-md... done
rpm-md repo 'updates' (cached); generated: 2026-01-06T01:08:16Z solvables: 18438
rpm-md repo 'fedora' (cached); generated: 2025-10-23T03:37:20Z solvables: 77664
rpm-md repo 'fedora-multimedia' (cached); generated: 2026-01-03T10:30:30Z solvables: 420
rpm-md repo 'updates-archive' (cached); generated: 2026-01-06T02:00:40Z solvables: 21271
Resolving dependencies... done
Checking out packages... done
error: Checkout fedora-logos-42.0.1-2.fc43.noarch: Hardlinking fd/57c11f7dbc00b368b7e6af4e7d259436736fd7af9aa3fa08aae8e420f3b3a2.file to fedora-gdm-logo.png: File exists
Looking for updates…

Nothing to do.
✔︎ JSON API cask.jws.json                                                                                         Downloaded   15.3MB/ 15.3MB
✔︎ JSON API formula.jws.json                                                                                      Downloaded   32.0MB/ 32.0MB
`

### What did you expect to happen?

No errors during ujust update

### Output of `bootc status`

```shell
❯ sudo bootc status
● Booted ostree
           Commit: 8bac411f547d2de9b95ce13451313f7dd9704d63d8cbc97b11b4fa468f8e2e6f

  Rollback ostree
           Commit: 47ffe50ba59308773db8c6538656fa8fbe2022afc629648bc330152cce2b658b
```

### Output of `groups`

```shell
❯ groups
wouter wheel dialout docker incus-admin libvirt
```

### Extra information or context

ghcr.io/ublue-os/bluefin-dx-nvidia-open:stable

renner0e:
[...]

PR #4072, opened 2026-01-18
fix: do not remove fedora-logos

tulilirockz:
We already override them with the package copying... Probably not necessary

Fixes: https://github.com/ublue-os/bluefin/issues/3999
Fixes: https://github.com/ublue-os/bluefin/issues/4048
Fixes: https://github.com/projectbluefin/iso/issues/21

NOTE: NEEDS TESTING

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -96,25 +96,31 @@ jobs:
 
       - name: Setup Syft
         id: setup-syft
-        if: false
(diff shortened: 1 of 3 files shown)


renner0e:
fedora-logos provides way more stuff than what we replace it with, we will 100% miss some things.

renner0e:
How bad of an idea would it be to query the files of the package and then rm -f chungus, but keep the package in the rpmdb this way

renner0e:
https://github.com/ublue-os/aurora/pull/1657
Did something for aurora

castrojo (approved):

Issue #3871, opened 2025-12-19
Interfering starship install and initialization


mheuvel-dev:
### Describe the bug

As a happy user of bluefin-cli, I am investigating the initialization sequence of starship and atuin. The atuin issue is bothering me ;)

I think there is a problem with the way starship is set up.

At the bottom of `.bashrc`, we source  `bling.sh` and looking inside it, there it is mentioned:

```
# Initialize atuin before starship to ensure proper command history capture
# See: https://github.com/atuinsh/atuin/issues/2804
```

However, on top of the `.bashrc`, there the `/etc/bashrc` is sourced and that file already initializes starship.

Looking in the bluefin git repository, it seems that starship is already installed via a tar package from github before the `/etc/bashrc` is populated, see [build_files/base/05-override-install.sh](https://github.com/ublue-os/bluefin/blob/76b562708c12c8c069dfcc5a24b4d96397b24877/build_files/base/05-override-install.sh#L23)

Isn't the brew install of starship preferred?
I think that starship is considered part of the default experience, regardless of bluefin-cli. However, now it interferes with the bluefin-cli initialization. Any suggestions how to move forward?

Happy to help further.

### What did you expect to happen?

I expected that starship is installed only once and initialized  in the sequence as mentioned in `bling.sh`.

### Output of `bootc status`

```shell
● Booted image: ghcr.io/ublue-os/bluefin-dx:stable
        Digest: sha256:21860f73f7a8e44b613e5eb29ad86ceacdedb552a0a7cc46f9f7a0ca1d3c6f92 (amd64)
       Version: 43.20251216 (2025-12-16T01:45:47Z)
```

### Output of `groups`

```shell
<user> wheel docker incus-admin libvirt
```

castrojo:
Thanks for working on this! I'll let others comment on the other parts, but for this:

> Isn't the brew install of starship preferred?
> I think that starship is considered part of the default experience, regardless of bluefin-cli. However, now it interferes with the bluefin-cli initialization. Any suggestions how to move forward?

This is an artifact from a time where we didn't have homebrew in Bluefin, so we added it as part of the image. In later versions of Bluefin, like LTS, we get it from brew and ship a stock `user@hostname` prompt. 

Ideally we'd do starship from brew everywhere across the board, however it's probably worse to remove starship from the image from thousands of people who are already used to it. We're working on freshening this up so it should get much cleaner moving forward. 

[...]

PR #4147, opened 2026-02-01
fix: move starship init bash to `/etc/profile.d`

salim-b:
Copied over from https://github.com/get-aurora-dev/common/pull/110.

Fixes https://github.com/ublue-os/bluefin/issues/3871.


salim-b:
@castrojo I just noticed that `starship init bash` is also included in `/usr/share/ublue-os/bling/bling.sh` from https://github.com/projectbluefin/common:

https://github.com/projectbluefin/common/blob/857a1d22d418e8af8c57959ccf8ad5e2fcb7942a/system_files/shared/usr/share/ublue-os/bling/bling.sh#L50

I currently don't use/source `/usr/share/ublue-os/bling/bling.sh`, so I'm unaffected by this duplication, I guess.

Should this PR's [`/etc/profile.d/90-bluefin-starship.sh`](https://github.com/ublue-os/bluefin/pull/4147/changes#diff-8ccf4e796756778e40b142ab204f9ac388c9c6e28e1c4a83a217661a313924e7) file better be moved to https://github.com/projectbluefin/common (and `starship init bash` removed from `bling.sh`?)

agriffis:
Just a heads up, as noted in https://github.com/ublue-os/bluefin/discussions/4163, this change causes starship to be reactivated for users who have previously deactivated it according to the FAQ.

PR #4211, opened 2026-02-19
feat: rewrite changelog generator to use SBOMs instead of rechunker

hanthor:
This PR rewrites the changelog generator to use `skopeo` and `cosign` for fetching package lists and SBOMs directly from the container registry.

Changes:
- Replaces legacy manifest inspection with `skopeo inspect` and `cosign verify-attestation`.
- Parses SPDX JSON SBOMs to extract package versions.
- Improved performance with threaded fetching.
- Updates `Justfile` to use the new script arguments.
- Supports automatic tag discovery.

### DIFF
--- a/.github/workflows/generate-release.yml
+++ b/.github/workflows/generate-release.yml
@@ -36,6 +36,12 @@ jobs:
         with:
           fetch-depth: 500
 
+      - name: Install cosign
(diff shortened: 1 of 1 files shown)


castrojo:
Is it possible to make this an action out of the gate? It'd be nice to decouple the changelogs from the image entirely so everyone can just use one action and it integrates with them, then custom images can have all of this ootb.

hanthor:
Yeah working on it https://github.com/hanthor/changelog-action

Issue #4263, opened 2026-03-07
Bazaar simply stops working occasionally. It happened to me 3 times. I only installed Bluefin 3 days ago.


kalebepalacio:
### Describe the bug

Sometimes when I go to use the computer and need to open the bazaar, it simply doesn't open. Other times it opens normally. 
I'm sending two screenshots, one of the system log and one of the image status showing that I haven't made any layering changes (even though I know Baazar is a flatpak).

The message in the log about Bazaar is as follows: Failed to start bazaar.service - Bazaar background service.

<img width="1250" height="650" alt="Image" src="https://github.com/user-attachments/assets/d3b7f721-2195-43f4-a016-ab5a130c0dec" />
<img width="1034" height="657" alt="Image" src="https://github.com/user-attachments/assets/130e6ade-eb76-4360-ac44-7fa5996ff989" />

### What did you expect to happen?

The bazaar should open normally and I should be able to install new applications.

### Output of `bootc status`

```shell
❯ sudo bootc status
[sudo] senha para kalebepalacio: 
● Booted image: ghcr.io/ublue-os/bluefin:stable
        Digest: sha256:2253e0cae7f80e942a53e5220eaddee535ad2f2ce8c14bf1534bba6353a31c53 (amd64)
       Version: 43.20260303 (2026-03-03T01:45:41Z)

  Rollback image: ghcr.io/ublue-os/bluefin:stable
          Digest: sha256:2253e0cae7f80e942a53e5220eaddee535ad2f2ce8c14bf1534bba6353a31c53 (amd64)
         Version: 43.20260303 (2026-03-03T01:45:41Z)
```

### Output of `groups`

```shell
❯ groups
kalebepalacio wheel
```

kalebepalacio:
As I mentioned before, there are times when it simply starts working again and opens up.

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/58391855-af65-4aa7-970f-ac80cb2b2bb9" />

PR #4264, opened 2026-03-07
fix: turn off bazaar.service

renner0e:
This has been an ever-growing pain in the past few weeks especially. Users sometimes were hit with random timeouts that no one could ever figure out the cause for. This means that search results for bazaar will only appear in the shell when it has been explicitly started by the user beforehand.

Best to disable it and leave the option to enable it if desired.

We can revisit this later when we have found a better solution to this. Aurora has done this since a couple weeks and hasn't received a single report since then.

Fixes: https://github.com/ublue-os/bluefin/issues/4263
Possibly Related: https://github.com/kolunmi/bazaar/issues/1005

See: get-aurora-dev/common#113

### DIFF
--- a/build_files/base/17-cleanup.sh
+++ b/build_files/base/17-cleanup.sh
@@ -8,7 +8,7 @@ set -eoux pipefail
 sed -i 's|uupd|& --disable-module-distrobox|' /usr/lib/systemd/system/uupd.service
 
 # Setup Systemd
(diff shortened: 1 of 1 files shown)


tulilirockz:
Could we also delete the service later then?

renner0e:
> Could we also delete the service later then?

yup of course we can delete the service file later

PR #4355, opened 2026-03-28
feat(beta): turn on Bluefin beta

castrojo:
Tada!

Thanks to Aurora for going first!

https://github.com/castrojo/bluefin/pkgs/container/bluefin/763654164?tag=beta

To inspect the content: 

```
skopeo inspect docker://ghcr.io/castrojo/bluefin:beta \
  | python3 -c "import json,sys; l=json.load(sys.stdin)['Labels']; print(l['org.opencontainers.image.version']); print(l['ostree.linux'])"
```

### DIFF
--- a/build_files/base/20-tests.sh
+++ b/build_files/base/20-tests.sh
@@ -34,22 +34,6 @@ flatpak preinstall --help
 # Make sure this garbage never makes it to an image
 test -f /usr/lib/systemd/system/flatpak-add-fedora-repos.service && false
 
-# Fedora flatpak remotes must never be present — flathub only
-# The fedora remote pollutes the app store with unmaintained/duplicate apps
-if flatpak remote-list --system | grep -q "^fedora"; then
-    echo "ERROR: fedora flatpak remote is present — this must not ship in the image"
-    exit 1
-fi
-if flatpak remote-list --system | grep -q "^fedora-testing"; then
-    echo "ERROR: fedora-testing flatpak remote is present — this must not ship in the image"
-    exit 1
-fi
-# Flathub must be present
-if ! flatpak remote-list --system | grep -q "^flathub"; then
-    echo "ERROR: flathub flatpak remote is missing"
-    exit 1
-fi
-
 IMPORTANT_PACKAGES=(
     distrobox
     fish


castrojo:
Oh, can someone make a `beta` branch in the repo?

build_files/base/04-packages.sh
```diff
@@ -157,6 +156,9 @@ case "$FEDORA_MAJOR_VERSION" in
     43)
         EXCLUDED_PACKAGES+=(gnome-software cosign)
         ;;
+    44)
+        EXCLUDED_PACKAGES+=(gnome-software cosign)
+        ;;
```
renner0e: I don't understand why we are removing gnome-software for every fedora version version separately. This must be leftovers when GTS used to not have bazaar/have GS on the image.

Justfile
```diff
@@ -557,12 +560,31 @@ fedora_version image="bluefin" tag="latest" flavor="main" $kernel_pin="":
...
+
+        # Verify akmods existence
+        echo "Checking for akmods: ghcr.io/ublue-os/akmods:${AKMODS_FLAVOR}-${VERSION}" >&2
+        if ! skopeo inspect --retry-times 3 "docker://ghcr.io/ublue-os/akmods:${AKMODS_FLAVOR}-${VERSION}" > /dev/null; then
+             echo "Error: Akmods not found for flavor ${AKMODS_FLAVOR} and version ${VERSION}" >&2
+             exit 1
+        fi
+
+        # Write cache
+        echo "{\"Labels\": {\"org.opencontainers.image.version\": \"${VERSION}\"}}" > /tmp/manifest.json
```
renner0e: we removed this in aurora, it doesn't make any sense (for our purposes), https://github.com/ublue-os/aurora/pull/1966

you are still based on ublue-os/silverblue-main and use skopeo copy for the akmods images, I don't think this makes sense for you as well

build_files/base/20-tests.sh
```diff
@@ -34,6 +34,22 @@ flatpak preinstall --help
...
+if flatpak remote-list --system | grep -q "^fedora-testing"; then
+    echo "ERROR: fedora-testing flatpak remote is present — this must not ship in the image"
+    exit 1
+fi
+# Flathub must be present
+if ! flatpak remote-list --system | grep -q "^flathub"; then
+    echo "ERROR: flathub flatpak remote is missing"
+    exit 1
+fi
+
```
renner0e: I don't think this makes any sense, the flatpak remote commands operate on /var/lib/flatpak, which you are not/cannot ship in a bootc image anyway.

Protecting yourself from fedora flatpaks means not shipping `/usr/lib/systemd/system/flatpak-add-fedora-repos.service` which is part of the flatpak package in fedora. This runs on first boot.


Additionally you should also turn this service off, this won't ever trigger because the associated unit is gone, but it will be leftover in the filesystem for nothing. 
`/etc/systemd/system/multi-user.target.wants/flatpak-add-fedora-repos.service`
https://src.fedoraproject.org/rpms/fedora-release/blob/f44/f/90-default.preset#_376

Which is handled by https://github.com/ublue-os/aurora/pull/1879/changes/309a1182c585cfef23463f820bb2247140078ae6

PR #4376, opened 2026-03-30
fix: clear all versionlocks in clean-stage.sh

castrojo:
Add `dnf versionlock clear` after `dnf clean all` to ensure kernel versionlocks added during build do not persist into the final image. Mirrors Aurora PR 1817.

Assisted-by: Claude Sonnet 4.6 via GitHub Copilot


castrojo:
Closing — `dnf versionlock clear` is already present in `build_files/shared/clean-stage.sh` in current main. This change landed as part of a prior merge. Nothing to do here.

PR #4393, opened 2026-04-01
chore: remove ARG fallbacks

dylanmtaylor:
For consistency with Aurora


dylanmtaylor:
Local build passes through F44 base image, kernel, and akmods installation. Fails on a pre-existing appindicatorsupport GNOME extension schema issue already present on the beta branch.

castrojo (approved):
**Code review — LGTM**

Removes hardcoded ARG defaults (`"silverblue"`, `"42"`, `"6.10.10-200.fc40.x86_64"`) from Containerfile for Aurora/bluefin parity. The `just build` recipe already passes all args explicitly via `BUILD_ARGS`, so removing defaults has no effect on normal builds. Prevents silent use of stale fallback values if args are accidentally omitted.

Reviewed by castrojo via pi.

PR #4504, opened 2026-04-21
fix: exclude gnome-software on all Fedora versions

dylanmtaylor:
Moves gnome-software to the common EXCLUDED_PACKAGES list so it is removed on F44+. F42/43 explicitly exclude the gnome-software package, but F44 doesn't. We're not building F41 anymore, so we can just move to common excludes.


dylanmtaylor:
I'm not sure if cosign should be moved to the common EXCLUDE_PACKAGE list too. It's only needed by CI to sign images.

castrojo (approved):

hanthor (approved):

PR #4539, opened 2026-04-28
fix: ensure xdg-desktop-portal starts after gnome-keyring-daemon

leafyoung:
Pika Backup cannot connect to the secret service to retrieve or save passwords after each upgrade.

## Reason

xdg-desktop-portal may start before gnome-keyring-daemon registers org.freedesktop.portal.Secret, leaving the Secret portal interface unavailable. This causes Pika Backup and other apps relying on the Secret portal to fail with "background process inactive" or keyring access errors.

## Solution

Add a systemd user drop-in that orders xdg-desktop-portal after gnome-keyring-daemon, ensuring the Secret backend is always discovered during portal startup.

### DIFF
--- a/system_files/shared/usr/lib/systemd/user/xdg-desktop-portal.service.d/30-after-keyring.conf
+++ b/system_files/shared/usr/lib/systemd/user/xdg-desktop-portal.service.d/30-after-keyring.conf
@@ -1,0 +1,9 @@
+[Unit]
+# Ensure gnome-keyring-daemon is running before the portal starts,
+# so the Secret portal backend (org.freedesktop.portal.Secret) is
+# available when xdg-desktop-portal probes for backends.
+# Without this, the portal may start before the keyring and cache
+# stale backend discovery, causing apps like Pika Backup to fail
+# with "background process inactive" or keyring access errors.
+After=gnome-keyring-daemon.service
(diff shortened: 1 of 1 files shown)


castrojo:
Thanks for this! Reviewers, should this go here or in common?

castrojo:
## Lab Verification Report — ghost/KubeVirt (2026-05-23)

Tested in `titan-bluefin` VM (KubeVirt, ghost `192.168.1.102`, NodePort 30223) on `bluefin-dx:latest-44.20260523.1`.

**Drop-in file present:**
```
/usr/lib/systemd/user/xdg-desktop-portal.service.d/30-after-keyring.conf
-rw-r--r--. 1 root root 461 Jan 1 1970
```

**Drop-in content verified:**
```ini
[Unit]
After=gnome-keyring-daemon.service
Wants=gnome-keyring-daemon.service
```

**systemd ordering confirmed** (`systemctl --user cat xdg-desktop-portal.service`):
```
After=graphical-session.target
After=gnome-keyring-daemon.service
Wants=gnome-keyring-daemon.service
```

**PASS** — drop-in is present, content is correct, systemd sees the ordering.

_Tested by castrojo via pi_

castrojo (approved):
Lab verified — PASS. See test report above. Ready to merge.

PR #4664, opened 2026-05-22
ci: attach SBOMs to releases

hanthor:
## Summary
- upload each non-PR build SBOM as a workflow artifact from reusable-build
- download those SBOM artifacts in generate-release and attach them to the GitHub release
- grant release jobs actions: read so the artifact download step can run

## Alignment
- Parallel bluefin-lts PR: ublue-os/bluefin-lts#1360

## Validation
- just check
- pre-commit run --files .github/workflows/reusable-build.yml .github/workflows/generate-release.yml .github/workflows/build-image-stable.yml .github/workflows/build-image-latest-main.yml .github/workflows/build-image-beta.yml

### DIFF
--- a/.github/workflows/build-image-beta.yml
+++ b/.github/workflows/build-image-beta.yml
@@ -45,6 +45,7 @@ jobs:
     needs: [build-image-beta]
     permissions:
       contents: write
+      actions: read
     secrets: inherit
     uses: ./.github/workflows/generate-release.yml
     with:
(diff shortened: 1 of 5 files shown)


renner0e:
Why?

If we want to do this then I think we should enable immutable releases, sign those artifacts and provide a way to verify them as part of the release notes? Maybe we could also mention how to actually do something useful with them.

castrojo (approved):

castrojo:
That seems like a fine idea to me? I'm using the SBOMs to keep the docs and website up to date, wouldn't hurt to just have immutable releases with them attached on each one anyway?

`oras discover ghcr.io/ublue-os/bluefin:stable` also works?

castrojo:

PR #4682, opened 2026-05-26
fix: remove wallpaper build rewrite

joshyorko:
## Summary

- Remove the build-time wallpaper month rewrite from `05-override-install.sh`.

## Why

Dynamic wallpaper rotation is moving to the shared common layer in projectbluefin/common#349. Keeping this build-time sed block would still mutate the default wallpaper gschema during Bluefin image builds.

This is the Bluefin-side cleanup for the approved dynamic wallpaper work from ublue-os/bluefin#3890.

## Testing

- `bash -n build_files/base/05-override-install.sh`
- `git diff --check HEAD~1..HEAD`

Refs projectbluefin/common#349
Refs ublue-os/bluefin#3890


castrojo:

PR #4762, opened 2026-06-14
fix: repo command for fwupd swap in 04-packages.sh

castrojo:
Should fix the build.


renner0e (commented):
You can probably rip this out completely, as long as ID is set to fedora in os-release.

And the secureboot thing from Microsoft is getting updated soon so if something doesn't work then people will absolutely complain and we get our confirmation this way :D

https://github.com/ublue-os/aurora/pull/2273

Issue #4764, opened 2026-06-16
[GDX] initramfs 345MB causes GRUB "out of memory" on 1GB /boot partition


RaimondoMorosini:
**Describe the bug**
On bluefin-gdx:lts (stream10.1, 2026-04-07), the initramfs for kernel 
6.19.7-100.fc42.x86_64 is 345MB. On systems with the standard 1GB /boot 
partition, GRUB fails to load it with:

  error: ../../grub-core/kern/mm.c:552:out of memory.

followed by kernel panic: VFS: Unable to mount root fs on unknown-block(0,0)

The previous deployment (stream10.1, 2025-09-16, kernel 6.12.0) had a 
244MB initramfs and booted correctly.

**Root cause**
The GDX initramfs includes all GPU drivers (nvidia, amdgpu, radeon, i915, 
xe, nouveau) + ZFS uncompressed, totaling 345MB. On a 1GB /boot partition 
with two deployments, this leaves no room for GRUB to operate.

  /boot size:   959MB
  initramfs:    345MB  ← too large
  vmlinuz:       18MB
  /boot free:   287MB  ← insufficient for GRUB

**Proposed fix**
Add a dracut config in the GDX Containerfile:

  RUN echo 'compress="zstd -19 -T0"' \
      > /usr/lib/dracut/dracut.conf.d/99-gdx-slim.conf && \
      echo 'omit_drivers+=" amdgpu radeon i915 xe "' \
      >> /usr/lib/dracut/dracut.conf.d/99-gdx-slim.conf

GDX targets Nvidia+CUDA workloads — AMD/Intel GPU drivers in the initramfs 
are unnecessary and add ~130MB. With zstd compression and nvidia-only 
drivers, the initramfs drops from 345MB to ~157MB (tested locally).

**Workaround (tested)**
Manually regenerate the initramfs after boot from ostree:1:

  sudo dracut \
    --force \
    --compress "zstd -19 -T0" \
    --omit-drivers "amdgpu radeon nouveau i915 xe" \
    --omit "fips" \
    --kmoddir /sysroot/ostree/deploy/default/deploy/<hash>/usr/lib/modules/<kver> \
    --kver <kver> \
    /boot/ostree/<path>/initramfs-<kver>.img

**Environment**
- Image: ghcr.io/ublue-os/bluefin-gdx:lts
- Version: stream10.1 (2026-04-07)
- Kernel: 6.19.7-100.fc42.x86_64
[...]

PR #4767, opened 2026-06-16
fix: slim initramfs for nvidia systems

bashilias:
fix for https://github.com/ublue-os/bluefin/issues/4764

draacut drop in for nvidia images that enables zstd compression and skips drivers that arent needed for nvidia systems.


bashilias:
messy pr, my badd ill create a new one

Issue #3354, opened 2025-10-10
Kernel panic/GRUB out of memory when booting 24.20251005.1


LoganM94:
### Describe the bug

When my system attempts to boot to Bluefin 24.20251005.1, the message below appears:

`error: ../../grub-core/kern/mm.c:552:out of memory.`

`Press any key to continue...`

When pressing the any key, the kernel panic message appears.

`KERNEL PANIC!`

`Please reboot your computer.`

`VFS: Unable to mount root fs on unknown-block(0,0)`

### What did you expect to happen?

To boot into Bluefin 24.20251005.1 normally.

### Output of `bootc status`

```shell
● Booted image: ghcr.io/ublue-os/bluefin:stable
        Digest: sha256:02cdd6bf7890c05f08498840e7694b4bf6a0d066c1372a770d4895833e0fae97 (amd64)
       Version: 42.20250928.1 (2025-09-28T06:02:29Z)
        Pinned: yes

  Rollback image: ghcr.io/ublue-os/bluefin:stable
          Digest: sha256:1fa3bdc7dd11ee242ab5ad8776292c87544b55793fc42f74ebe9f2cfa6fffe07 (amd64)
         Version: 42.20251005.1 (2025-10-05T06:02:51Z)
```

### Output of `groups`

```shell
logan wheel dialout docker incus-admin lxd libvirt
```

### Extra information or context

I've also attempted to rebase to bluefin:stable-daily and bluefin:dx, with the same issue after staging/reboot.

I can boot back into my current image (42.20250928.1) normally.

I've attempted to switch OFF Secure Boot, but the issue persists. Secure Boot has been switched back on.

L0wgain:
Heya, disabling **Intel Platform Trust Technology** in UEFI lets me boot into 24.20251005.1.

Unsure if this should stay open, since this is a workaround and not a complete resolution.

mhetzi:
Hi
I too get this on a ThinkCentre M920q, but did not find any UEFI option to fix this.
The newer initramfs is 62,7MB bigger than the one that works.

<img width="1139" height="434" alt="Image" src="https://github.com/user-attachments/assets/7aba58ca-06e7-404a-9568-88a7ac59170e" />

laurikala:
I have troubles too on an Elitesk 800 G1 SFF machine. 

[...]

PR #4768, opened 2026-06-16
fix: slim initramfs for nvidia systems

bashilias:
fix for https://github.com/ublue-os/bluefin/issues/4764

draacut drop in for nvidia images that enables zstd compression and skips drivers that arent needed for nvidia systems.


bashilias:
compression is working, down to 214mb but amdgpu.ko.xz and i915.ko.xz are still in the initramfs despite omitting the drivers because 99-nvidia.conf uses force_drivers which overrides omit_drivers i believe going to get rid of the drivers with sed.

bashilias:
1. sed strips i915, amdgpu, radeon, xe, nouveau from force_drivers in 99-nvidia.conf
2. zz-nvidia-slim.conf (loads last, alphabetically after 99-nvidia.conf) adds omit_drivers for those same drivers and upgrades compression to zstd -19 -T0.

tested this and im getting initramfs of 179mb. Fix could probably be better but for now it won;t be a show stopper

castrojo:
How big is your boot? I'm wondering if we should set that to 2GB if we haven't already.

renner0e:
The issue you mentioned is a problem on bluefin lts, you are on the wrong repo, development has also moved to the projectbluefin org.

Correct me if I'm wrong but wouldn't this break amd/Intel systems that are on the Nvidia image?

People switching from Nvidia to amd and never rebasing to the proper image and systems on external drives that are supposed to be used on Nvidia and everything else comes to mind.

What about dual GPU systems?

bashilias:
valid concern, users with Nvidia gpu switching to amd without rebasing to the non-nvidia image will get a broken boot. . Maybe increasing the boot size is the better solution but existing installs with 1 GB /boot don't benefit from a partition size change, they'd still need a workaround. But that's a smaller population than the risk of breaking dual gpu users.

bashilias:
Another solution would be just adding compression instead of getting rid of the drivers.

renner0e:
> Another solution would be just adding compression instead of getting rid of the drivers.

How much of an actual difference does it make if you just crank the compression level to the max and nothing else. Any noticable speed differences?

bashilias:
tested compression only. It only saves a few mb  (<10mb) since the firmware blobs are already pre compressed. Driver removal is the only real fix (345mb > 179mb) for the size issue but raises valid dual GPU concerns. This compression can be implemented for now but as @castrojo said the boot also needs to be increased, however it won't help current installs. The -19 compression level only adds a bit of build time on CI, boot speed is unaffected since zstd decompression is the same at all levels.

mhetzi:
#3354 can you add a fix for non nvidia images too? because on my system the boot partition is 2GB.

The initramfs with all these firmware blobs is simply too big for some EFI systems.

image

Why is there nvidia firmware when there are no drivers in the intel/amd version

bashilias:
Thinking about this more, for dual gpu laptops (Nvidia + Intel iGPU), i915 isn't actually needed in the initramfs. The initramfs only runs long enough to mount the real rootfs, display is handled by the Nvidia driver via nvidia-drm.modeset=1. i915 loads from the real rootfs afterwards anyway.
The real question for the maintainers is; what is the nvidia image actually targeting?

- Dedicated Nvidia workstations ? driver removal is safe, 
- Also Nvidia+Intel hybrid machines?

We cna keep the driver removal fix, its probably safe even for hybrid GPU systems, remove only AMD drivers (amdgpu/radeon),  and keep i915/xe for hybrid laptop safety or bump /boot to 2GB in installer, but @mhetzi  shows even 2GB isn't always enough..

Long term; might have to think about per-variant initramfs so each image only carries its own GPU firmware.

renner0e:
> Also Nvidia+Intel hybrid machines?

yes, dual GPU systems like laptops are of course supposed to work

bashilias:
The root cause is hostonly=no which is set deliberately for atomic images, meaning all gpu firmware gets bundled regardless of variant. Compression doesn't help since the firmware blobs are already pre compressed individually. Even increasing /boot won't scale as firmware keeps growing with new gpu generations. The only sustainable fix is per variant initramfs where each image only ships firmware for its own gpu family. 

For the nvidia image this means removing amdgpu, radeon and nouveau but keeping i915 and xe for Nvidia+Intel hybrid laptops. For the non-nvidia image it means removing nvidia firmware which would fix the issue @mhetzi raised. Each image ends up carrying only what it needs. Switching gpus would require rebasing to the correct image which is already the expected workflow in the atomic model. Want to go this route?

bashilias:
i updated the fix to also remove nvidia firmware from non nvidia images. results; nvidia image 327MB to 194MB (kept i915/xe for hybrid laptops), non nvidia image 236MB to 134MB. This should fix the issues raised by @mhetzi and @renner0e . 
existing installs wont benefit automatically, they need to rebase or regenerate initramfs manually.

castrojo:
Ok here's an idea, we have projectbluefin/bluefin:stable and :testing going now, why don't we just put it in testing, this let's people opt in and you can just let it rip, we're migrating people there eventually anyway. 

Might be a good time to get fresh ideas in too?

castrojo:
CLosing this, will make sure we have this in the new bluefin

PR #4819, opened 2026-07-06
fix(gnome): use gsettings override for blur-my-shell

castrojo:
## Summary
- Move the Blur my Shell popup default to a system-wide GSettings override
- Remove the per-user dconf write from the theming setup hook

## Testing
- `bash -n system_files/shared/usr/share/ublue-os/user-setup.hooks.d/10-theming.sh`


castrojo:
For reviewers, the projectbluefin/common repo has moved on so we need to patch it here before migration.

ahmedadan (approved):

PR #4886, opened 2026-08-07
fix(build): actually pin the base image by digest and unblock Renovate

castrojo:
## Why

`image-versions.yml` had a `silverblue-main` digest that **nothing consumed**:

- `Justfile` only read the `common` and `brew` digests via `yq`.
- `Containerfile` built `FROM ${BASE_IMAGE}:${FEDORA_MAJOR_VERSION}` - a floating tag.

So the base image was never actually pinned, and every Renovate bump of that field for months was a no-op. Bumping the value alone would not have changed a single byte of the built image.

Separately, Renovate had not opened a base image PR in a month. #4832 and #4878 were sitting in **"PR Edited (Blocked)"** on dependency dashboard #2680, because `main` had been hand-merged into their `renovate/*` branches three times. Renovate will not update *or* replace a branch it no longer owns, and the config had `rebaseWhen: "never"`, which makes hand-merging the only way to freshen a stale branch. The PRs looked green and mergeable the whole time.

## What

- Wire `BASE_IMAGE_SHA` through `just build` into the `FROM`, so the pin is real and cosign verifies the exact digest being built.
- Bump the base digest to current `latest`, and bump `brew`.
- `rebaseWhen: "never"` -> `"behind-base-branch"` so Renovate maintains its own branches.

Pins are keyed by **Fedora major version**, not by tag stream. `quay.io/fedora/fedora-coreos:stable` currently reports 44, the same as `latest` - so keying on the stream would have pulled an F43 base into a build whose akmods and kernel resolved for F44. Keying on the version `just build` already resolved makes that mismatch structurally impossible.

At a Fedora rollover the lookup fails loudly with `No digest pinned for silverblue-main-45 ... Add an entry for Fedora 45 before building.` rather than silently falling back.

## Scope

Trimmed after review. Dropped from an earlier revision, as unrelated to the fix: an `AGENTS.md` documentation section, a corrected action version comment, a Renovate rule disabling `ublue-os/remove-unwanted-software`, an explicit `image-versions.yml` custom manager, and the stale `FEDORA_MAJOR_VERSION` ARG defaults.

The custom manager in particular was verified unnecessary: aurora has no such manager yet its dashboard still detects `image-versions.yml`, and Renovate already produced #4832/#4878 from this file. Adding one risked duplicate detection.

Net: **31 insertions, 9 deletions across 4 files.**

## Note for the reviewer

The existing `automerge: true` digest rule for `ghcr.io/ublue-os/silverblue-main` is unchanged, but it is now **load-bearing** rather than cosmetic: base image digest bumps will auto-merge and will now genuinely change the built image. Left as-is deliberately - flagging in case you want that revisited.

## Aurora

Aurora has the identical latent bug: its `kinoite` digest in `image-versions.yml` is dead data, and `Containerfile.in` uses the floating tag. It likely wants the same fix.

## Verification

- `just check`, `renovate-config-validator --strict`, and `pre-commit` on the changed files all pass. (`.devcontainer.json` and `.github/ISSUE_TEMPLATE/config.yml` fail pre-commit on `main` already.)
- Cosign-verified all three digests.
- Simulated the lookup for F43, F44 and F45.

The two red `generate-release` checks are pre-existing repo-wide and are **not** required by the `main` ruleset; unrelated PR #4884 fails on the identical digest. Root cause is fixed separately in #4887.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,8 +8,10 @@
     "enabled": true
   },
 
-  // Do not rebase when the default branch is updated
-  "rebaseWhen": "never",
+  // Renovate cannot touch a branch that has been manually edited, which
+  // silently stalls the dependency for months. Keep its own branches current
+  // instead of hand-merging main into them.
+  "rebaseWhen": "behind-base-branch",
     customManagers: [
     {
       customType: 'regex',
--- a/Containerfile
+++ b/Containerfile
@@ -2,6 +2,7 @@ ARG BASE_IMAGE_NAME="silverblue"
 ARG FEDORA_MAJOR_VERSION="42"
 ARG SOURCE_IMAGE="${BASE_IMAGE_NAME}-main"
 ARG BASE_IMAGE="ghcr.io/ublue-os/${SOURCE_IMAGE}"
+ARG BASE_IMAGE_SHA=""
 ARG COMMON_IMAGE="ghcr.io/projectbluefin/common:latest"
 ARG COMMON_IMAGE_SHA=""
 ARG BREW_IMAGE="ghcr.io/ublue-os/brew:latest"
@@ -19,7 +20,7 @@ COPY --from=brew /system_files /system_files/shared
 COPY /system_files /system_files
 
 ## bluefin image section
-FROM ${BASE_IMAGE}:${FEDORA_MAJOR_VERSION} AS base
+FROM ${BASE_IMAGE}:${FEDORA_MAJOR_VERSION}@${BASE_IMAGE_SHA} AS base
 
 ARG AKMODS_FLAVOR="coreos-stable"
 ARG BASE_IMAGE_NAME="silverblue"
--- a/Justfile
+++ b/Justfile
@@ -135,8 +135,18 @@ build $image="bluefin" $tag="latest" $flavor="main" rechunk="0" ghcr="0" pipelin
     fi
     fedora_version=$({{ just }} fedora_version '{{ image }}' '{{ tag }}' '{{ flavor }}' '{{ kernel_pin }}')
 
-    # Verify Base Image with cosign
-    {{ just }} verify-container "${base_image_name}-main:${fedora_version}"
+    # Base image digest pin, keyed by the resolved Fedora version so the pinned
+    # digest can never disagree with the version everything else is built for.
+    base_image_entry="${base_image_name}-main-${fedora_version}"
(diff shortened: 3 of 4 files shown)


castrojo:
## CI status: all 16 build jobs pass; the 2 failures are pre-existing and unrelated

**Every image build passed** - bluefin and bluefin-dx, main and nvidia-open, latest and stable. That is the code path this PR changes, so the digest pinning works end to end.

The two red checks are `Generate Release / generate-release` (`latest` and `stable-daily`), both failing with:

```
RuntimeError: No SBOM referrer found for ghcr.io/ublue-os/bluefin@sha256:00278a5e...
```

This is not caused by this PR:

- **Control test:** PR #4884 (an unrelated `ossf/scorecard-action` bump) fails with the *identical error on the identical digest* `00278a5e`. It is repo-wide, affecting any PR opened today.
- **This PR publishes nothing.** SBOM generation, image push, and SBOM upload in `reusable-build.yml` are all gated on `if: github.event_name != 'pull_request'`, so a PR build cannot create an image that lacks an SBOM.
- **`changelogs.py` only reads already-published images.** `get_manifests()` runs `skopeo inspect` against `ghcr.io/ublue-os/bluefin:<tag>` and diffs the last two published tags. It never touches the PR build.
- **Confirmed at the registry:** the GHCR referrers API returns `0` referrers for `00278a5e`, so the published `stable-daily-20260807` image genuinely has no SPDX attachment. `latest-20260807` (`ffaf4f31`) is in the same state.

So today's main-branch publish produced images without SBOM referrers, and every PR now trips over it during changelog generation. That is a separate bug on `main` and should be fixed independently - this PR should not be blocked on it.

castrojo:
Root cause of the two red `generate-release` checks is now understood and fixed separately in #4887: the push step gave each tag its own `podman push`, so every tag got a different manifest digest and only the default tag's digest received the signature, SBOM and attestation.

Those checks are **not** required by the `main` ruleset (only the two build checks are, and both pass here), so this PR is not gated on them. Auto-merge is armed; it needs one approving review.

castrojo:
Trimmed after a review pass against ublue-os/aurora. Removed the AGENTS.md section, the action version comment fix, the `remove-unwanted-software` Renovate rule, the explicit `image-versions.yml` custom manager, and the stale ARG default changes. Now 31 insertions / 9 deletions across 4 files, down from 73/12 across 6.

The custom manager was the notable cut: aurora detects its `image-versions.yml` with no such manager, so adding one here was redundant and risked duplicate detection.

